PR #4, opened 2016-05-09
fix for typo in MANIFEST.in

serein7:


LeonChambers:
Oops, thanks for catching this. I just pushed the fix for it.

PR #14, opened 2016-05-26
Add environment info endpoint

gordonbrander:
@LeonChambers can you review and suggest edits?

### DIFF
--- a/openag/brain/core/run_modules.py
+++ b/openag/brain/core/run_modules.py
@@ -14,6 +14,7 @@ from .models import ModuleModel, ModuleTypeModel, ModuleConnectionModel
 from .db_names import DbName
 from .db_server import db_server
 from .parameters import Parameter
+from .var_types import EnvironmentalVariable
 
 def main():
     # Parse command line arguments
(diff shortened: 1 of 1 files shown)


openag/brain/core/run_modules.py
```diff
@@ -99,10 +100,57 @@ def main():
...
     app.mod.start()
 
+
+    def find_recipe_start(env_data_db, env_id):
+        """Given a pointer to the environmental data point db, return the
+        latest recipe start for a given environment.
+        returns a Dict or None.
+        """
+        # @TODO this constant should exist somewhere else
+        recipe_start_type = 'recipe_start'
```
LeonChambers: This constant is in `openag.brain.core.var_types`

openag/brain/core/run_modules.py
```diff
@@ -99,10 +100,57 @@ def main():
...
+
+        return recipe_handler_ids[0] if len(recipe_handler_ids) > 0 else None
+
+
     @app.route("/<mod_id>/<endpoint>", methods=['POST'])
     def serve_endpoint(mod_id, endpoint):
         return getattr(app.mod.ask(mod_id), endpoint)(**request.json)
 
+
+    @app.route("/environment/<env_id>", methods=['GET'])
```
LeonChambers: If there is a module named `environment`, it will be impossible to call endpoints on it because queries to `/environment/<endpoint>` will go to this route instead of the endpoint one.

openag/brain/core/run_modules.py
```diff
@@ -99,10 +100,57 @@ def main():
     app.mod = Module(mod_id)
     app.mod.start()
 
+
+    def find_recipe_start(env_data_db, env_id):
```
LeonChambers: You should check if the recipe has ended already before returning it.

openag/brain/core/run_modules.py
gordonbrander: Fixed by adding API versioning and namespacing.

openag/brain/core/run_modules.py
gordonbrander: :+1: fixed

gordonbrander:
@LeonChambers all issues addressed. Can you review + land?

PR #15, opened 2016-06-06
Add cors support through flask-cors

gordonbrander:


gordonbrander:
@LeonChambers can you review/land?

Issue #35, opened 2016-09-01
Pillow missing from Docker image


gordonbrander:
Moved from https://github.com/OpenAgInitiative/openag_brain_docker_rpi/issues/10.

PR #36, opened 2016-09-01
Install pillow without zlib and jpeg

gordonbrander:
Fix for https://github.com/OpenAgInitiative/openag_brain/issues/35.

@LeonChambers can you review and provide feedback re: whether we want zlib and jpeg?


LeonChambers:
I'm pretty sure this would be fine. Right now, it just reads the raw images and writes them as PNGs, so I doubt it uses zlib or jpeg. You could set up a camera and see if anything breaks if you want to test it.

PR #39, opened 2016-09-11
CORS removed

serein7:
Based on [this discussion](https://github.com/OpenAgInitiative/openag_brain_docker_rpi/pull/12) openag_brain api is supposed to be run through CouchDb proxy which already has CORS enabled. So this one has to be removed to avoid error with multiple values set 
>The 'Access-Control-Allow-Origin' header contains multiple values 'http://localhost, http://localhost', but only one is allowed

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -51,7 +51,7 @@
   <run_depend>python-requests</run_depend>
   <run_depend>python-gevent</run_depend>
   <run_depend>python-flask</run_depend>
-  <run_depend>python-flask-cors-pip</run_depend>
+
   <run_depend>python-lxml</run_depend>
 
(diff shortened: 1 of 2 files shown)


gordonbrander:
Thanks for catching this! We also need to remove the flask-cors dependency, if it is no longer needed https://github.com/OpenAgInitiative/openag_brain/blob/master/package.xml#L54.

gordonbrander:
@serein7 can you also remove the flask CORS dependency in this patch (see link above)? Or if you don't want to, I can fork the patch and do it.

serein7:
OK. I'll do later today

serein7:
@gordonbrander, removed

Issue #41, opened 2016-09-14
Can not add PID software_module


serein7:
@LeonChambers, I'm trying to setup feedback loop on light_illuminance. Based on my understanding of PID fixture in software_module_types the only required parameter is variable. So I add following into software_module:

`{
   "_id": "light_pid",
   "type": "openag_brain:pid.py",
   "parameters": {
       "variable": "light_illuminance"
   }
}`

My assumption that, nevertheless I did not specify other params like Kd, Ki, Kp and etc this module should still be wired correctly and run, even if PID will not produce valid outputs. 
However, I see following error in logs:

>2016-09-14T19:18:33.367723613Z Processing module "light_pid" from server
2016-09-14T19:18:33.366726437Z Traceback (most recent call last):
2016-09-14T19:18:33.367859705Z   File "/home/pi/catkin_ws/src/openag_brain/scripts/main", line 117, in <module>
2016-09-14T19:18:33.367948610Z     fixtures=vals.fixtures, screen=vals.screen
2016-09-14T19:18:33.368021057Z   File "/home/pi/catkin_ws/src/openag_brain/scripts/main", line 58, in main
2016-09-14T19:18:33.368152306Z     commands.update_launch(server)
2016-09-14T19:18:33.368564228Z   File "/home/pi/catkin_ws/src/openag_brain/src/openag_brain/commands/update_launch.py", line 99, in update_launch
2016-09-14T19:18:33.368667091Z     module_type = SoftwareModuleType(module_types_db[module["type"]])
2016-09-14T19:18:33.371513256Z   File "/usr/local/lib/python2.7/dist-packages/voluptuous/schema_builder.py", line 192, in __call__
2016-09-14T19:18:33.371671327Z     return self._compiled([], data)
2016-09-14T19:18:33.371751066Z   File "/usr/local/lib/python2.7/dist-packages/voluptuous/schema_builder.py", line 486, in validate_dict
2016-09-14T19:18:33.371827107Z     return base_validate(path, iteritems(data), out)
2016-09-14T19:18:33.371911272Z   File "/usr/local/lib/python2.7/dist-packages/voluptuous/schema_builder.py", line 324, in validate_mapping
2016-09-14T19:18:33.371987522Z     raise er.MultipleInvalid(errors)
2016-09-14T19:18:33.372056740Z voluptuous.error.MultipleInvalid: not a valid value for dictionary value @ data[u'parameters'][u'Kd'][u'type']

Any idea?

[...]

PR #42, opened 2016-09-15
added param type mapping from float to double

serein7:


gordonbrander:
@LeonChambers can you review?

@serein7 can you give a bit of background on this one? It looks important!

serein7:
@gordonbrander, here is the background https://github.com/OpenAgInitiative/openag_brain/issues/41

PR #47, opened 2016-09-19
Fix #46 (bad reference to master)

### LINKED ISSUE #46: Bug in get_topic_info endpoint
Variable `master` does not exist in scope. We want `rostopic_master`. See https://github.com/OpenAgInitiative/openag_brain/blob/master/src/openag_brain/software_modules/api.py#L175

gordonbrander:
Fixes https://github.com/OpenAgInitiative/openag_brain/issues/46.


gordonbrander:
@LeonChambers r?

PR #53, opened 2016-09-23
Return valid json responses from topic post api

gordonbrander:


gordonbrander:
Tested on Raspberry Pi/Arduino using Pin 13 LED as actuator.

PR #66, opened 2016-10-18
Change -f flag to use absolute filepaths for fixtures

gordonbrander:
## Rationale

Makes it easier to build custom Docker images from rpi brain by providing an absolute reference to your own 3rd party fixture file.

```Dockerfile
FROM openag/rpi_brain

USER pi
# Get the fixture files
RUN cd ~ && git clone http://github.com/me/my_fixture.git
CMD ["~/catkin_ws/devel/env.sh", "rosrun", "openag_brain", "main", "-f", "~/my_fixture/fixture.json"]
```

## Usage

Old:

    rosrun openag_brain main -f default

New:

    rosrun openag_brain main -f ~/path/to/default.json

## Note

Before merging, we should make a corresponding change to `docker-compose.yml` in https://github.com/OpenAgInitiative/openag_brain_docker_rpi.

### DIFF
--- a/scripts/main
+++ b/scripts/main
@@ -88,6 +88,17 @@ def main(api_server, db_server, fixtures, screen):
             print "Spawning software modules"
             modules = spawn_modules(screen)
 
+def resolve_fixtures(fixtures):
+    """
+    Given a list of fixture names, returns a list of fully resolved fixture
+    paths.
+    """
+    rospack = rospkg.RosPack()
+    pkg_path = rospack.get_path('openag_brain')
+    fixtures_path = os.path.join(pkg_path, "fixtures")
+    resolved = [os.path.join(fixtures_path, fixture_name + ".json") for fixture_name in fixtures]
+    return resolved
+
 if __name__ == '__main__':
     parser = argparse.ArgumentParser(
         description="""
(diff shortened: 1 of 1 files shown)


gordonbrander:
Hmm, while this works when openag_brain outside of Docker, I have not successfully gotten it to work within a Docker container.

LeonChambers:
I see the use in being able to load a fixture from an absolute path, but I think we should keep the ability to load the fixtures installed with `openag_brain` easily.

Also, the Dockerfile for the custom image could just load the fixture itself using `openag db load_fixture`.

LeonChambers:
Wait, nvm. The Dockerfile can't load the fixture itself because that requires the CouchDB container to be running. Maybe just having one flag for loading a fixture from the fixtures folder and a different flag for loading a fixture from an arbitrary path could work?

gordonbrander:
> 

Wait, nvm. The Dockerfile can't load the fixture itself because that requires the CouchDB container to be running. Maybe just having one flag for loading a fixture from the fixtures folder and a different flag for loading a fixture from an arbitrary path could work?

Yeah that was exactly the problem I ran into. Having 2 flags wfm. Will update the PR.

gordonbrander:
@LeonChambers added a separate flag `--fixture-path`. What do you think?

gordonbrander:
Tested new flag on Raspberry Pi 3, as well as confirming old flag still works.

PR #71, opened 2016-10-21
Binary direct controller

### LINKED ISSUE #70: direct_controller.py can only handle Float64 type messages
`direct_controller.py` [can only work with Float64 messages](https://github.com/OpenAgInitiative/openag_brain/blob/master/src/openag_brain/software_modules/direct_controller.py#L18).

We should have an analog to `direct_controller.py` that works with binary values so we can use it with `binary_actuator.py`.

gordonbrander:
Fixes #70

### DIFF
--- a/fixtures/default.json
+++ b/fixtures/default.json
@@ -518,6 +518,34 @@
       }
     },
     {
+      "_id": "openag_brain:on_off_controller.py",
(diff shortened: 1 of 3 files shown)


gordonbrander:
@LeonChambers review?

Issue #52, opened 2016-09-23
start_recipe API is broken


serein7:
The latest commit which has it working is 88b01ee
Than starting from dc25676cd1d34c87b635ca050e89d2f2ddbf7d4f whenever I try to call recipe_start I get following exception:

>Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/gevent/pywsgi.py", line 508, in handle_one_response
    self.run_application()
  File "/usr/lib/python2.7/dist-packages/gevent/pywsgi.py", line 494, in run_application
    self.result = self.application(self.environ, self.start_response)
  File "/usr/lib/python2.7/dist-packages/flask/app.py", line 1836, in __call__
    return self.wsgi_app(environ, start_response)
  File "/usr/lib/python2.7/dist-packages/flask/app.py", line 1820, in wsgi_app
    response = self.make_response(self.handle_exception(e))
  File "/usr/lib/python2.7/dist-packages/flask/app.py", line 1403, in handle_exception
    reraise(exc_type, exc_value, tb)
  File "/usr/lib/python2.7/dist-packages/flask/app.py", line 1817, in wsgi_app
    response = self.full_dispatch_request()
  File "/usr/lib/python2.7/dist-packages/flask/app.py", line 1477, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/usr/lib/python2.7/dist-packages/flask/app.py", line 1381, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/usr/lib/python2.7/dist-packages/flask/app.py", line 1475, in full_dispatch_request
    rv = self.dispatch_request()
  File "/usr/lib/python2.7/dist-packages/flask/app.py", line 1461, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/home/pi/catkin_ws/src/openag_brain/src/openag_brain/software_modules/api.py", line 127, in perform_service_call
    res = rosservice.call_service(service_name, [args])[1]
  File "/opt/ros/indigo/lib/python2.7/dist-packages/rosservice/__init__.py", line 411, in call_service
    rospy.init_node('rosservice', anonymous=True)
  File "/opt/ros/indigo/lib/python2.7/dist-packages/rospy/client.py", line 271, in init_node
    raise rospy.exceptions.ROSException("rospy.init_node() has already been called with different arguments: "+str(_init_node_args))
[...]

PR #73, opened 2016-10-24
Fix init_node bug in api by using ServiceProxy

gordonbrander:
We previously used rosservice.call_service to call ROS services. It
turns out this is an internal API and does spooky unwanted things like
initializing a node from within the function. This is what caused
https://github.com/OpenAgInitiative/openag_brain/issues/52 and we
thought we had fixed it with https://github.com/OpenAgInitiative/openag_brain/pull/64/files#diff-d5fc26493f28134f9baefcdbeefdcf91L40.

Unfortunately, that didn't actually fix the problem. Removing init_node meant
that rosservice.call_service worked, but nothing else did (for example
publishing to a rostopic won't work if you don't initialize your node).

This fixes the issue by using the public API, ServiceProxy.

Unfortunately, ServiceProxy has a limitation that you must import the ROS
ServiceClass, which is autogenerated from srv files. This makes it difficult
to do a generalized endpoint for services, since we pass in the service name,
but don't know the name of the service class from the HTTP call.

This commit provides a fix by creating 2 static endpoints: /start_recipe and
/stop_recipe and removes the generalized POST API for services. It's not ideal,
but it works.

### DIFF
--- a/src/openag_brain/software_modules/api.py
+++ b/src/openag_brain/software_modules/api.py
@@ -18,9 +18,6 @@ from flask import Flask, jsonify, request, Response
 from gevent.wsgi import WSGIServer
 from gevent.queue import Queue
 
-from openag_brain.services import START_RECIPE, STOP_RECIPE
-from openag_brain.srv import StartRecipe, Empty
-
 API_VER = "0.0.1"
 
 app = Flask(__name__)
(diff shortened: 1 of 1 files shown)


gordonbrander:
@LeonChambers can you review. Plz also notice the tradeoff that I had to make.

gordonbrander:
Note this is a breaking change for recipe start API and will require UI to be updated.

LeonChambers:
I definitely don't like the switch away from a general API, but having working endpoints for starting and stopping recipes is better than having a broken general endpoint, so that could be fine for now. However, we should be able to switch to these specific endpoints without introducing a breaking change. In particular, if we name the endpoints like `/service/<environment>/start_recipe` (which is what they were before), we won't have to update the UI now for these endpoints and again later if we potentially reintroduce the general API.

There's definitely a way to figure out the name of the service class to use given the name of the service. The command line tool for calling a service uses [this function](https://github.com/strawlab/ros_comm/blob/master/tools/rosservice/src/rosservice.py#L346) for that. Ideally, we could just use this helper function to write the more general endpoint.

gordonbrander:
@LeonChambers keeping the same URL works for me. Initially, I thought it would be better not to expose under `/service` since we couldn't expose the full service API to POST. However, given this lookup function, that sounds like a good idea.

I'll update the PR to use the old URL, then open a follow-up issue for making this general again.

gordonbrander:
@LeonChambers ok, I skipped straight to making the service endpoint dynamic again. Note that json `args` is now required to be a JSON array, rather than a single value. This allows us to invoke services which may have > 1 arg.

Tested pretty carefully, but LMK if you see anything missed.

gordonbrander:
@LeonChambers gonna merge this so we have something that works... Please comment on any follow-up you might want.

LeonChambers:
LGTM

Issue #49, opened 2016-09-19
Replace resolve_message_type with roslib.message.get_message_class


gordonbrander:
I think we can replace `resolve_message_type` with the `roslib` function `get_message_class`. They both take a message type string and return the Python class for that message type.

``` python
from roslib.message import get_message_class
type_string = "std_msgs/Float32"
MsgType = get_message_class(type_string)
```

gordonbrander:
Here's the source for the roslib function https://github.com/ros/ros/blob/kinetic-devel/core/roslib/src/roslib/message.py#L95

LeonChambers:
Resolved in #74

PR #74, opened 2016-10-25
Replace resolve_message_type with roslib.message.get_message_class #49

yeliztaneroglu:


src/openag_brain/software_modules/topic_filter.py
```diff
@@ -8,6 +8,7 @@
 from couchdb import Server
 
 from openag_brain.util import resolve_message_type
```
gordonbrander: We should remove resolve_message_type if no longer used anywhere.

src/openag_brain/software_modules/sensor_persistence.py
```diff
@@ -27,6 +27,7 @@
 
 from openag_brain import params
 from openag_brain.util import resolve_message_type
```
gordonbrander: We should remove resolve_message_type if not used. (would also be good to check that it is unused elsewhere in the codebase + remove from utils)

yeliztaneroglu:
Okey, I removed resolve_message_type if not used. Thank you so much :+1:  @gordonbrander

src/openag_brain/software_modules/topic_connector.py
```diff
@@ -25,6 +25,7 @@
 from openag_brain import params
 from openag_brain.srv import Empty
 from openag_brain.util import resolve_message_type
```
gordonbrander: This import should be removed. We should also remove the `resolve_message_type` function definitition from `openag_brain.util` if it's no longer being used anywhere.

Thanks again for your contribs! Super excited to have you join the community. :)

gordonbrander:
Going to test this out today. Assuming it works, with the small changes above it looks great!

src/openag_brain/software_modules/topic_connector.py
yeliztaneroglu: I am so excited, too : ) yaay \o/

yeliztaneroglu:
I removed resolve_message_type in topic_connector.py, Thank you so much for your time :)

yeliztaneroglu:
review ?  @gordonbrander

gordonbrander:
Looking at it today :D

yeliztaneroglu:
awesome :smile:

gordonbrander:
Tested. It works.!

PR #81, opened 2016-11-11
Flash script

gordonbrander:
This adds a new command 'flash` that allows users to flash the Arduino manually. Example:

    rosrun openag_brain flash --dir ~/build

`--dir` is optional, and a tmp directory will be generated if none is provided.

### DIFF
--- a/scripts/flash
+++ b/scripts/flash
@@ -2,52 +2,19 @@
 import subprocess
 import argparse
 import tempfile
-import select
-
-# def log(x):
-#     print x
-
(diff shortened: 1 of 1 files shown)


LeonChambers:
In the case where a temporary directory is generated, it might make sense to delete that directory after flashing since there's basically no way the files would be used again

gordonbrander:
@LeonChambers I thought the same, but after looking into it, discovered that the Pi (and most Linux distros) have a cleanup routine for the `tmp` dir. Anything in it will be deleted next reboot.

PR #84, opened 2016-11-18
Add initial v2 fixture

gordonbrander:


serein7:
I see that PID controller finally appeared to control air_temperature! Cool. 
Does that meant that  current brain code wires all the topics for sensors and controllers automatically?

gordonbrander:
@serein7 yeah! `topic_connector.py` implicitly connects loops via the `variable` configured in the controller, sensor and actuator.

PR #85, opened 2016-11-18
Fix throttling bug where fast sensors stomp others

gordonbrander:
This fixes a bug where fast updating sensors would prevent other
environmental data points from getting stored. This was because
last_time and last_value fields were shared between all sensor types. We
were throttling by time and value delta, but were not comparing like to
like. A fast updating sensor would mean the delta for other sensors would
always fall under the min update threshold.

This fix makes sure throttling compares like to like, comparing each value
against a previously generated EnvironmentalDataPoint, which is indexed by
environment, is_desired, and variable.


gordonbrander:
Nope! I misdiagnosed the problem.

PR #89, opened 2016-12-12
Loosely couple nodes using standardized, statically-knowable topic names

gordonbrander:
The goal of this PR is to loosely couple nodes to each other, by introducing a standardized interface for environment topics. This allows nodes to integrate into the system without knowing anything about the system except the topic name they wish to deal with.

## Rational

Currently, nodes need to know a lot of configuration details about the firmware module system in order to participate. This is because topic names and connections are constructed by reading firmware module configuration from the database.

This tightly couples all nodes to the details of the firmware module configuration. If, for example, the firmware module configuration format changes in future, it would cause a breaking change to all nodes, because all nodes rely on these details in order to communicate.

This can be easily fixed by tweaking the architecture to have a standardized topic interface. Instead of reading database config details to construct topic names, use the list of supported environmental variables to construct topic names. The list of supported variables is standardized so this list can be known without having to read it from the database module config at runtime. Publishers and subscribers can read the list of environmental variables from `openag.var_types` and use these as a normalized topic interface.

The only places that need to know or care about the DB are those that actually need to work with the DB:

- `topic_connector.py` (which reads firmware config in order to connect firmware topics with environmental variable topics)
- Persistence modules (`sensor_persistence.py`, `image_persistence.py` which need the DB to store info)

All other nodes can participate using the standard environmental variable interface.

## Changes Introduced

This PR introduces a handful of changes to make connecting nodes by naming convention easier:

- Environmental variable topic names have changed slightly
  - Old: `/<env_id>/<modifier>/<env_var>`
  - New: `/environments/<id>/<env_var>/<modifier>`
  - Environment topics are prefixed with `/environments/<id>` to make it easier to find topics by environment.
- Persistence nodes must be instantiated per-environment (they get their environment ID by looking at the parent namespace). This has the fringe benefit of letting us turn persistence on/off per environment.
- `topic_filter.py` must be instantiated per-environment. Same reasons/fringe benefits as above.
- Topics are filtered at the environment leavel (rather than sensor topic level)
  - Old: `/sensors/<id>/raw -> /sensors/<id>/filtered -> /<env_id/measured/<env_var>`.
  - New: `/sensors/<id>/raw -> /environments/<env_id>/<env_var>/raw -> /environments/<env_id>/<env_var>/measured`

## What do topic names look like now?

Not too different. The major changes are the revised environment topic names:

```
/actuators/arduino_led/cmd
/actuators/blue_light_1/cmd
/actuators/cooler_1/cmd
/actuators/heater_1/cmd
/actuators/humidifier/cmd
/actuators/pump_1/cmd
/actuators/pump_2/cmd
/actuators/pump_3/cmd
/actuators/pump_4/cmd
/actuators/pump_5/cmd
/actuators/red_light_1/cmd
/actuators/white_light_1/cmd
/diagnostics
/environments/environment_1/air_carbon_dioxide/measured
/environments/environment_1/air_carbon_dioxide/raw
/environments/environment_1/air_humidity/measured
/environments/environment_1/air_humidity/raw
/environments/environment_1/air_temperature/commanded
/environments/environment_1/air_temperature/desired
/environments/environment_1/air_temperature/measured
/environments/environment_1/air_temperature/raw
/environments/environment_1/light_illuminance/commanded
/environments/environment_1/light_illuminance/measured
/environments/environment_1/light_illuminance/raw
/environments/environment_1/water_dissolved_oxygen/measured
/environments/environment_1/water_dissolved_oxygen/raw
/environments/environment_1/water_electrical_conductivity/measured
/environments/environment_1/water_electrical_conductivity/raw
/environments/environment_1/water_oxidation_reduction_potential/measured
/environments/environment_1/water_oxidation_reduction_potential/raw
/environments/environment_1/water_potential_hydrogen/measured
/environments/environment_1/water_potential_hydrogen/raw
/environments/environment_1/water_temperature/measured
/environments/environment_1/water_temperature/raw
/rosout
/rosout_agg
/sensors/am2315_1/air_humidity/raw
/sensors/am2315_1/air_temperature/raw
/sensors/atlas_ph_1/water_potential_hydrogen/raw
/sensors/mhz16_1/air_carbon_dioxide/raw
```

## Feedback

One change that could use feedback: I've created a new file `openag.env_vars` that contains lists of environmental variables paired with topic types. This makes it easy to support more than floats in `topic_connector.py` in future. The reason I didn't add this type information to the class is because I figured we wanted to keep `openag_python` decoupled from ROS.

I also introduced the concept of "environmental variable categories" (image and sensor). This is kind of a useful abstraction, since we use sensors to close loops, but cameras are a bit different.


gordonbrander:
@LeonChambers can you take a look at the above?

LeonChambers:
I guess having a `/environments` namespace makes sense even if only to prevent collisions in the case where someone creates an environment with the id `sensors` or `actuators` for whatever reason

I'm pretty indifferent to whether persistence nodes should be global or per-environment; either one could make sense. Not sure being able to turn off persistence per-environment is even useful, but sure.

For filtering, it might make sense for the filtering nodes to operate at a sensor level and read from the DB so that they can set filter weights based on parameters of the particular sensor being filtered like repeatability and accuracy. That was the main reasoning behind the current structure.

In general, this change will lead to some nodes (e.g. topic_filter) listening on topics for which there is no data since it doesn't check the DB to see whether or not there is an actual sensor for a given variable type. This definitely adds some overhead, but it also makes the code nicer. In practice, I don't think the overhead would actually be significant, though.

gordonbrander:
> For filtering, it might make sense for the filtering nodes to operate at a sensor level and read from the DB so that they can set filter weights based on parameters of the particular sensor being filtered like repeatability and accuracy. That was the main reasoning behind the current structure.

That makes sense.

> In general, this change will lead to some nodes (e.g. topic_filter) listening on topics for which there is no data since it doesn't check the DB to see whether or not there is an actual sensor for a given variable type. This definitely adds some overhead, but it also makes the code nicer. In practice, I don't think the overhead would actually be significant, though.

Yup understood. However, the list of topics is pretty short, and my understanding is that ROS does next to no work with subscribers until someone actually publishes to that channel.

LeonChambers:
I'd be fine with merging this. We could revert the filtering code later if we actually want to use info from the database, but I agree that this PR generally makes the architecture a little nicer.

PR #90, opened 2016-12-13
Fix memory leak by removing view queries in main

gordonbrander:
Diagnosis: when db reaches a certain size, creating the view cache takes quite a bit of time. At a certain point, that time exceeds 5s. However, we were requesting that view caches be rebuilt every 5s. If the cache takes longer than 5s to build, this ends up causing a memory leak, as the previous cache rebuild request is continually stomped by the next request.

This pull request fixes the issue by removing cache rebuild requests. Consumers can request cache rebuild themselves.

### DIFF
--- a/scripts/main
+++ b/scripts/main
@@ -72,10 +72,6 @@ def main(api_server, db_server, fixtures, screen):
     )["last_seq"]
     while True:
         time.sleep(5)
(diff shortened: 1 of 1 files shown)


gordonbrander:
Merging this. It's a tested, critical fix.

Issue #92, opened 2016-12-19
recipe_handler.py continues to publish setpoints after recipe ends


gordonbrander:
What happens:

- After recipe ends, the last setpoint is continually published to desired ROS topics

What should happen:

- After recipe ends, no further desired messages should be sent by `recipe_handler.py`.

LeonChambers:
This is actually on purpose. There's no sense of when a recipe actually "ends", so after the last set point, it's not clear when to stop publishing desired variable values, so I just had it keep publishing indefinitely. A better solution would be to update the recipe structure to have a way of indicating when the recipe is actually over.

PR #93, opened 2017-01-10
Refactor recipe handler to fix bugs

gordonbrander:
This PR fixes several bugs discovered with the current recipe handler.

- Recipes now post desired setpoints to db
- When recipes are resumed, recipe will fast-forward through all state
  changes to ensure correct state when resuming. Previoulsy recipe would
  resume without replaying prior state changes, causing bugs when
  recipes were resumed.
- Recipes now have a definite end point, so we can post a recipe_end and
  stop spamming setpoints after recipe has ended.
- Make recipe start/stop threadsafe via Rlock (service handlers are run in
  separate threads by ROS). Previously it was not threadsafe, and
  multiple service calls may have created undefined behavior.
- Simpler control flow

This is a significant refactor that reworks and simplifies control flow across processes.

- Recipe is a blocking iterator that yields a setpoint every 1s
- Recipes are run in the main process. The main process also handles cancelling recipes (stopping iteration), posting to DB and posting to ROS topics
- services are run in separate threads by ROS (this was the case previously). The set recipes by using RLock for synchronizing state with main process.

Tested:

- Starting recipe
- Echoing rostopic
- Checked database for correct saves
- Stopping recipe
- Recipe interrupted by shutdown
- Recipe resumed after restart


src/openag_brain/software_modules/recipe_handler.py
```diff
@@ -12,248 +12,276 @@
...
+from openag_brain.memoize import memoize
+from openag_brain.multidispatch import multidispatch
 
-class Recipe(object):
-    def __init__(self, _id, operations, start_time):
-        self._id = _id
-        self.operations = operations
-        self.start_time = start_time
+# Create a tuple constant of valid environmental variables
+VALID_VARIABLES = tuple(EnvVar.items.keys())
```
LeonChambers: Since this tuple is only used for lookups, using a set instead would be more efficient. Or a frozenset if you want something immutable.

src/openag_brain/software_modules/recipe_handler.py
```diff
@@ -12,248 +12,276 @@
...
-        "simple": SimpleRecipe
-    }
+class RecipeIdleError(Exception):
+    """Thrown when trying to clear a recipe, but recipe is already clear."""
+    pass
 
-    def __init__(self, server):
+class RecipeHandler:
+    def __init__(self, server, environment):
+        # We create a lock to ensure threadsafety, since service handlers are
```
LeonChambers: I'm pretty sure this lock isn't actually necessary in CPython because of the GIL, but I'm also fine with leaving this in.

LeonChambers:
I mentioned this in #92, but recipes currently have no way of specifying when they actually end. The implementation checked into master continues publishing set points from a recipe indefinitely, and this implementation stops publishing set points when the last line of the recipe is encountered. Neither of these is really correct. The best solution I can think of is to have recipes actually contain a dummy recipe_end set point to signal the end of the recipe. This would literally only change 1 line of the logic in this PR and some of the documentation of the recipe specification, but I think it's important for correctness.

src/openag_brain/software_modules/recipe_handler.py
gordonbrander: I wasn't sure if ROS ran services in processes (independent of GIL) or threads (dependent on GIL), so this might just be paranoia. I would like to keep this for correctness anyway, if that's fine with you.

gordonbrander:
@LeonChambers in this PR, recipes are responsible for signaling recipe start and recipe end by yielding a start and end setpoint in order to resolve this issue (see https://github.com/OpenAgInitiative/openag_brain/pull/93/files?diff=split#diff-97ee90a90e020a51e597a5efe508da0fR106)

LeonChambers:
No, I meant in the recipe itself, not in the recipe interpreter. For example, with the way things are now, if I wanted to create a recipe that just keeps the temperature at 25 degrees for a day and then ends, the way I write that as `[[0, "air_temperature", 25]]`. When your code interprets that recipe, it would emit a desired air temperature value once, then emit a recipe_end and stop immediately. When I'm suggesting is that we change the documentation [here](http://openag-python.readthedocs.io/en/latest/recipes.html) to say that the recipe should be written as `[[0, "air_temperature", 25], [86400, "recipe_end", None]]` so that the set point is actually enforced for as long as the recipe author wanted it to be. This would mean that the recipe interpreter would no longer have to explicitly emit a recipe_end datapoint because the specs would dictate that that be in the recipe itself, but in retrospect, it's fine for a recipe_end to be emitted twice because it wouldn't get written to the database anyway, and it's safer in case a recipe author doesn't completely follow the specs.

So... that means my only problem with this PR is a potential documentation bug in another repo, so I'm fine with merging this now.

gordonbrander:
Ah makes sense. I'll fix the bitrot and commit.

PR #95, opened 2017-01-10
Updates to video_writer for creating timelapses

LeonChambers:
This PR contains a significant re-write of the video_writer, which creates timelapses from the aerial images taken from a camera in the environment. Timelapses are now generated in a way that correctly reflects the real time between images and should scale better than the previous method.

Currently, the variable being listened to for camera images is hard-coded to "aerial_image". It would be nice to generate a timelapse for every camera, but that would require a list of all of the variable types that can contain image data. This is added in a `var_types` module in #89, which hasn't been merged yet, so I'm waiting on that.

I have tested this a little locally for short time-scales, but ROS seems to have trouble reading from my camera for some reason, so the results a a bit hard to interpret. This needs to be tested on the real hardware at longer time-scales before being merged. Some of the operations (particularly the video conversion) might have to be optimized a little more before this is deployable.


src/openag_brain/software_modules/video_writer.py
```diff
@@ -4,21 +4,72 @@
...
+TIMELAPSE_ATTACHMENT = "timelapse"
+
+IMAGE_FILENAME = "image.png"
+WORKING_TIMELAPSE_FILENAME = "timelapse.mpeg"
+TIMELAPSE_FILENAME = "timelapse.mp4"
+SNIPPET_FILENAME = "snippet.mpeg"
+
+# By default, each second of the timelapse will correspond to 1 day in real
+# time
+DEFAULT_TIMELAPSE_SCALING_FACTOR = 24*60*60
```
gordonbrander: Nit: the name of this variable made me think it was for dimensional scaling (x,y) of video file.

gordonbrander:
I did a quick sanity check, by running this branch on the Pi. I can't test it further because I don't have a camera handy.

Looking through the code, it LGTM. We should test further if we can, but I'm fine with merging this now. Can always disable by removing from config.

PR #98, opened 2017-01-13
Add support for phased recipe format

gordonbrander:
Also:

- Fixes a bug with the interpret_registry dispatch (dispatch on "format" key, not "type" key).
- Fixes a bug with topic name (dispatch to `var/desired` not `desired/var`.

Phased recipes have the following format:

```json
{
   "_id": "test_short",
   "_rev": "1-86cba5eba9c01c5839bc4614dac35bb4",
   "format": "phased",
   "version": "1.0",
   "author": "openag",
   "recipe_format": "vector 1.0",
   "optimization": [
       "general purpose"
   ],
   "recipe_name": "default basil",
   "date_created": 1481616000,
   "recipe_id": "472309485702983470592837405",
   "operations": [
       {
           "cycles": 2,
           "name": "germination",
           "day": {
               "hours": 0.001
           },
           "night": {
               "light_intensity_blue": 0,
               "water_electrical_conductivity": 0,
               "water_potential_hydrogen": 6,
               "hours": 0.001,
               "air_temperature": 30,
               "air_humidity": 75,
               "light_intensity_white": 0,
               "air_carbon_dioxide": 1000,
               "light_intensity_red": 0
           }
       },
       {
           "cycles": 2,
           "name": "early",
           "day": {
               "hours": 0.001
           },
           "night": {
               "light_intensity_blue": 0,
               "water_electrical_conductivity": 0,
               "water_potential_hydrogen": 2,
               "hours": 0.001,
               "air_temperature": 40,
               "air_humidity": 80,
               "light_intensity_white": 0,
               "air_carbon_dioxide": 400,
               "light_intensity_red": 0
           }
       }
   ],
   "author_id": "457028937405687305698723059"
}
```

Tested on Raspberry Pi.

### DIFF
--- a/src/openag_brain/software_modules/recipe_handler.py
+++ b/src/openag_brain/software_modules/recipe_handler.py
@@ -105,6 +105,43 @@ class SimpleRecipe:
             yield (time.time(), variable, value)
         yield (time.time(), RECIPE_END.name, self.id)
 
+def hrs_to_seconds(hrs):
+    return hrs * (60 * 60)
+
+def days_to_seconds(days):
+    return hrs_to_seconds(days * 24)
+
+@interpret_recipe("phased")
+class PhasedRecipeInterpreter:
+    def __init__(self, recipe, start_time=None, timeout=1):
+        init_time = time()
+        start_time = start_time or init_time
+        if start_time > init_time:
+            raise ValueError("Recipes cannot be scheduled for the future")
+        self.start_time = start_time
+        self.stages = recipe["stages"]
+        self.id = recipe["_id"]
+        self.timeout = timeout
+
+    def __iter__(self):
+        yield (rospy.get_time(), RECIPE_START, self.id)
+        for stage in self.stages:
+            stage_duration = days_to_seconds(stage["days"])
+            end_of_stage = (rospy.get_time() - self.start_time) + stage_duration
+            phases = ("day", "night")
+            phase_count = 1
+            while (rospy.get_time() - self.start_time) < end_of_stage:
+                phase = stage[phases[phase_count % 2]]
+                phase_count += 1
+                phase_keys = frozenset(phase.keys())
+                phase_duration = hrs_to_seconds(phase["hours"])
+                end_of_phase = (rospy.get_time() - self.start_time) + phase_duration
+                while (rospy.get_time() - self.start_time) < end_of_phase:
+                    for key in VALID_VARIABLES.intersection(phase_keys):
+                        yield float(timestamp), key, float(phase[key])
+                    rospy.sleep(self.timeout)
+        yield (rospy.get_time(), RECIPE_END, self.id)
+
 class RecipeRunningError(Exception):
     """Thrown when trying to set a recipe, but recipe is already running."""
     pass
--- a/src/openag_brain/software_modules/recipe_handler.py
+++ b/src/openag_brain/software_modules/recipe_handler.py
@@ -38,7 +38,10 @@ def publisher_memo(topic, MsgType, queue_size):
     """
     return rospy.Publisher(topic, MsgType, queue_size=queue_size)
 
-@multidispatch(lambda x: x.get("type", "simple"))
+def get_format(recipe):
+    return recipe.get("format", "simple")
+
+@multidispatch(get_format)
 def interpret_recipe(recipe):
     """We don't support default behavior for recipe interpretation"""
     raise ValueError("Recipe type not supported")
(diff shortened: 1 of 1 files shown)


src/openag_brain/software_modules/recipe_handler.py
```diff
@@ -105,6 +108,43 @@ def __iter__(self):
...
+        self.start_time = start_time
+        self.stages = recipe["operations"]
+        self.id = recipe["_id"]
+        self.timeout = timeout
+
+    def __iter__(self):
+        yield (rospy.get_time(), RECIPE_START.name, self.id)
+        for stage in self.stages:
+            stage_duration = days_to_seconds(stage["days"])
+            end_of_stage = (rospy.get_time() - self.start_time) + stage_duration
```
LeonChambers: I'm probably just being pedantic here, but this computes the end time for the current stage relative to the current time instead of relative to when the previous stage was supposed to end. I believe the latter semantics would be more "correct". So basically I'm suggesting initializing `end_of_stage` to `self.start_time` outside of the loop and then adding `stage_duration` to it on every iteration. With the current implementation, stage start times are expected to drift forward from where you would naturally expect them to be by about `self.timeout/2` seconds per phase. This probably won't matter at all for most people tbh, but it's also a relatively easy change.

src/openag_brain/software_modules/recipe_handler.py
```diff
@@ -105,6 +108,43 @@ def __iter__(self):
...
+            phases = ("day", "night")
+            phase_count = 1
+            while (rospy.get_time() - self.start_time) < end_of_stage:
+                phase = stage[phases[phase_count % 2]]
+                phase_count += 1
+                phase_keys = frozenset(phase.keys())
+                phase_duration = hrs_to_seconds(phase["hours"])
+                end_of_phase = (rospy.get_time() - self.start_time) + phase_duration
```
LeonChambers: The above comment also applies here

src/openag_brain/software_modules/recipe_handler.py
gordonbrander: Good suggestion.

gordonbrander: @LeonChambers ah, I'm remembering why I wrote it this way.

Imagine you have a 19hr stage with a 12hr night and a 12hr day. What should happen?

You could enforce the stage duration and truncate the last phase:

     stage |-------------------|
     phase |------------|------|

Or you could allow the phase to run its course and overrun the stage time

     stage |-------------------|
     phase |------------|------------|

I prefer the latter, because you can imagine a situation where a night is cut short, followed by an intense day in the next stage and the plant ends up burned. So overrunning the stage seems like the most benign solution... a sensible default.

Of course, this implementation doesn't solve everything. One could also imagine the following mistake happening:

    |day|night|day|night|day|
                            |day|night|day|night|

But I didn't bother to guard against it. Maybe we should!

Bottom line, I think it makes more sense to let the stage time drift long if it means completing phases correctly, since I think this will be a better default for growth outcomes.

Thoughts?

LeonChambers: Ok, I see your point.

Well, we could just consider a recipe with a 19hr stage with a 12hr night and 12hr day to be an invalid recipe and not worry at all about what happens in that case. However, this means that a typo in recipe could often lead to dead plants, which isn't very nice.

Perhaps a better way to handle this is to re-design the recipe format in such a way that a situation like this couldn't happen at all. For example, instead of specifying that a stage last for a particular length of time, we could have the user specify that the stage last for some number of day-night cycles. To me, this seems slightly easier to reason about when writing recipes and prevents the possibility of mathematical errors resulting in dead plants.

gordonbrander: Yeah, I agree it's easier to reason about programmatically, but I think it's less intuitive when writing recipes.

For example, I know a basil takes about 20 days to grow: 3 days to germinate, 7 days to mature, and 10 more days until harvest. During maturity I know my day cycle can be 14hrs etc, etc. Jotting down that "farmer's knowledge" in a recipe is easy, and intuitive even if it's off by an hour, give or take. But converting that farmer's knowledge to number of cycles via back-of-the-napkin math is kind of a pain.

Conversely, if I'm being exact, I can always take out my pencil, do some long division and make sure everything lines up correctly.

That is my take anyway.

LeonChambers: I personally think it's much easier to reason about stages as a number of day-night cycles. And if, as we often claim, plants actually grow faster in our system than with other conventional methods, directly transcribing "farmer's knowledge" isn't particularly useful anyway.

Another way to handle this if we keep the current semantics is to have the system basically "check your math" and round the `days` field to the nearest value that is a multiple of the length of the day-night cycle.

gordonbrander: Upon reflection, I think I'm ok with having `days` become `day_cycles`, and be a multiplier for `day + night`. So `day_cycles: 5` would be valid.

Your argument brings me around: it's hard to reason about solar 24hr days overlayed against artificial day lengths, regardless. Better to reason about phases (you know the plant can handle a 14 hr day in this stage, etc) and how many cycles you want it to go through.

gordonbrander:
Addressed all comments in latest commit. Tested on Raspberry Pi.

gordonbrander:
@LeonChambers do you want to do a quick review on the result?

gordonbrander:
I'm merging this.

PR #99, opened 2017-01-15
Refactor controller modules and fix recipe_handler

davoclavo:
Hello! First of all, I've got to say this is an awesome project and I am excited to contribute. Here are some of my recent ideas:

- I think converting from one type to another should be done by the controller module. I added an `on_off_controller` re-implementation where I tried to show some of the nuances of defining a measurement+setpoint[Float64] -> commanded[Bool] transition. For example, for a thermostat with a boolean actuator using [hysteresis](https://en.wikipedia.org/wiki/Hysteresis#Control_systems).

- With that in mind, I added a Controller base class that can be inherited in controller classes to dynamically set up the environment such as loading default ROS Params, registering callbacks to set points and measurements ROS Topics, and publishing commanded output values.

- Fixes recipe_handler bug that caused the node to crash when the recipe was stopped by a ROS service call (via the UI or via `rosservice call`). Moved the methods to set/clear the ROS Parameters in the setter/getter as stopping could be called via the ROS Service. Also fix the desired topic to conform to the new format `/<variable>/desired`.

- I removed the default topics the Controllers were publishing and subscribing to, as I couldn't find any code that referenced them. I assumed it was cruft from a previous implementation, now the module raises an exception if no `"variable"` is set.

Extra thoughts:

- I noticed that `direct_controller.py` was meant to send the output directly from a recipe, but I think a more appropriate concept for this idea would be to use closed-loop vs open-loop controllers. I haven't implemented it, but I think it would require to attach the publisher to the set point callback, and do any type manipulation there and then publish, without needing a measured sample to update the output.
- I encourage feedback/questions


LeonChambers:
Hey,

First of all, thanks for contributing. It's great to see people interested in the project. This PR has a few good bug fixes, but some of the architectural changes will require some discussion.

1) In reference to the default topics for the controllers, in the `inputs` and `outputs` dictionary for a `firmware_module_type`, the keys are typically used directly as topic names. When the module is a controller, this no longer makes sense because we want the topic names to be determined at run time based on the value of the `variable` parameter. The way I chose to dealt with this was to have the modules use the actual keys as topic names by default and only change them if the `variable` parameter was set. Then, I was very explicit in the `firmware_module_type` record that the topic names might be different at run time. I actually used the default topics names a lot during testing so that I could run the module from the command line without having to set any ROS parameters, but you are correct that none of the code in this repo directly uses them. I'm not convinced that it is worth it to remove this default behavior, and we definitely need to be more explicit in the fixture that the topic names can change based on the runtime configuration of the module.

2) The `direct_controller` that was in the repo was a simple open-loop controller. We have been using this controller with the lights since we don't have sensor feedback. As you described, it just latches to the set point topic and passes the value through to the command topic. It also outputs a fake "measurement" making explicit in the ROS topic structure the assumption that the actuator actually achieves the commanded value. If you have any other ideas on how to handle open-loop controllers, I'd be glad to discuss them. However, this PR replaces the `direct_controller` with a closed-loop proportional controller, which is just not going to work for the use case we needed.

3) You renamed the file for the PID controller file without updating the `executable` value for it in the fixture, which will break things.

4) You removed the logic in `topic_connector` that uses the `multiplier` and `deadband` values. I guess this change goes along with your first point in this thread. I acknowledge that these parameters can be abused to do type conversion in the `topic_connector`, but this is not their purpose. Their purpose is to abstract hardware concerns from controller logic. For example, when configuring a temperature control loop, it is confusing to think about heating and cooling being provided by different pieces of hardware (a separate heater and chiller). A nice abstraction here is for there the controller to produce a single control effort value where sufficiently positive values cause the heater to turn on and sufficiently negative values cause the chiller to turn on. One of the 2 outputs might have to be scaled as well to reflect differences in the efficiencies of the 2 different pieces of the hardware. Such a configuration (which we plan to use on the v2 hardware) would require both the `multiplier` and `deadband` values. A agree that the controller should be aware of whether it's controlling a binary or continuous actuator because that has real implications for the control algorithm used, but there are some things that are better handled at lower layers.

I'll have to do a full code review before merging, but these are my initial thoughts after reading through everything.

davoclavo:
> In reference to the default topics for the controllers, in the inputs and outputs dictionary for a firmware_module_type, the keys are typically used directly as topic names. When the module is a controller, this no longer makes sense because we want the topic names to be determined at run time based on the value of the variable parameter. The way I chose to dealt with this was to have the modules use the actual keys as topic names by default and only change them if the variable parameter was set. Then, I was very explicit in the firmware_module_type record that the topic names might be different at run time. I actually used the default topics names a lot during testing so that I could run the module from the command line without having to set any ROS parameters, but you are correct that none of the code in this repo directly uses them. I'm not convinced that it is worth it to remove this default behavior, and we definitely need to be more explicit in the fixture that the topic names can change based on the runtime configuration of the module.

I feel like testing artifacts shouldn't be embedded in production code, and mocking the ROS environment would be better, but as currently there is no test suite, it makes a lot of sense to keep them there for manual testing. I'll revert my changes to add them back and remove my bit of code that caused an exception if no `~variable` param was set.

> The direct_controller that was in the repo was a simple open-loop controller. We have been using this controller with the lights since we don't have sensor feedback. As you described, it just latches to the set point topic and passes the value through to the command topic. It also outputs a fake "measurement" making explicit in the ROS topic structure the assumption that the actuator actually achieves the commanded value. If you have any other ideas on how to handle open-loop controllers, I'd be glad to discuss them. However, this PR replaces the direct_controller with a closed-loop proportional controller, which is just not going to work for the use case we needed.

Thanks for clarifying the usage of the direct controller (and open-loop in general), it wasn't that clear to me why the topics where connected that way. 
I guess the only reason to re-publish to `/measured` to persist those set points as environmental data points (and display them in the UI), but please let me know if there is something else I'm missing.
I'll implement a base class for open-loop controllers and re-add a compatible direct_controller logic for the use case without sensor..

> You renamed the file for the PID controller file without updating the executable value for it in the fixture, which will break things.

Good catch, I'll fix that.

> You removed the logic in topic_connector that uses the multiplier and deadband values. I guess this change goes along with your first point in this thread. I acknowledge that these parameters can be abused to do type conversion in the topic_connector, but this is not their purpose. Their purpose is to abstract hardware concerns from controller logic. For example, when configuring a temperature control loop, it is confusing to think about heating and cooling being provided by different pieces of hardware (a separate heater and chiller). A nice abstraction here is for there the controller to produce a single control effort value where sufficiently positive values cause the heater to turn on and sufficiently negative values cause the chiller to turn on. One of the 2 outputs might have to be scaled as well to reflect differences in the efficiencies of the 2 different pieces of the hardware. Such a configuration (which we plan to use on the v2 hardware) would require both the multiplier and deadband values. A agree that the controller should be aware of whether it's controlling a binary or continuous actuator because that has real implications for the control algorithm used, but there are some things that are better handled at lower layers.

Thanks for the explanation, clearly I wasn't considering that a single environmental variable could be controlled by more than one actuator. Looking at this [previous issue](https://github.com/OpenAgInitiative/openag_brain/issues/70) seems that this had already been discussed, I'll try to come up with a simple way to support Boolean types in the topic_connector.

fixtures/default.json
```diff
@@ -467,23 +467,23 @@
...
           "type": "std_msgs/Float64",
-          "description": "The message data element must contain the current value of the controlled plant property. The controller publishes std_msgs/Float64 messages on the control_effort topic each time it receives a message on the state topic. Thus the rate at which the plant publishes state governs the control-loop rate - the plant should publish state at the desired loop rate. This topic can be renamed by the topic_from_plant parameter"
+          "description": "The message data element must contain the current value of the controlled plant property. The controller publishes messages on the commanded topic each time it receives a message on the state topic. Thus the rate at which the plant publishes state governs the control-loop rate - the plant should publish state at the desired loop rate."
         }
       },
       "outputs": {
         "cmd": {
-          "type": "std_msgs/Float64",
-          "description": "The control_effort message data element contains the control effort to be applied to the process to drive state/data to equal setpoint/data. This topic can be renamed by the topic_from_controller parameter"
```
davoclavo: Couldn't find any references to `topic_from_controller` in the code, but let me know if this shouldn't be removed

LeonChambers: Oh, yea. I used the [ROS PID node](http://wiki.ros.org/pid) as a starting point for this node and then took out some of the parameters. So this should be removed

davoclavo:
Hey, here is my latest update:

I added a base class to implement open-loop controllers and closed-loop controllers with a simple interface. It sets up all the ROS related things like loading the params and the topic subscription/publishing.

Here are a couple of examples:

From `direct_controller.py` - using `OpenLoopController`
(note: the `set_point` is not part of the internal state, instead it is passed as argument to `update`)

```
from controllers import OpenLoopController

class Direct(OpenLoopController):
    def update(self, set_point):
        return set_point

if __name__ == '__main__':
    Direct.start()
```

from `pid_controller.py` - using `ClosedLoopController`

```
from controllers import ClosedLoopController

class PID(ClosedLoopController):
    """ Discrete PID control """

    def __init__(self, Kp=0, Ki=0, Kd=0, upper_limit=1, lower_limit=-1,
            windup_limit=1000, deadband_width=0):
        self.Kp = Kp
        self.Ki = Ki
        self.Kd = Kd
        self.upper_limit = upper_limit
        self.lower_limit = lower_limit
        self.windup_limit = windup_limit
        self.deadband_width = deadband_width

        self.last_error = 0
        self.integrator = 0

    def update(self, state):
        if self.set_point is None:
            return

        error = self.set_point - state

        if abs(error) < self.deadband_width:
            return 0

        p_value = self.Kp * error
        d_value = self.Kd * (error - self.last_error)
        self.last_error = error
        self.integrator = self.integrator + error
        self.integrator = max(-self.windup_limit, min(self.windup_limit, self.integrator))
        i_value = self.Ki * self.integrator

        res = p_value + i_value + d_value
        res = min(self.upper_limit, max(self.lower_limit, res))

        return res

if __name__ == '__main__':
    PID.start()
```

Also added support for booleans in the topic connector. In order to do so, I needed the configuration for the software module outputs, so I am fetching the output command type from the configured variable in the software_module db. 

This allows controller modules to define Bool as an output command. Here is an example of how to specify the `output_type` for a controller.

```
from controllers import OpenLoopController
from std_msgs.msg import Bool

class OnOffOpen(OpenLoopController):
    output_type = Bool

    def __init__(self, threshold=0, activate_below_threshold=False):
        self.threshold = threshold
        self.activate_below_threshold = activate_below_threshold

    def update(self, set_point):
        if self.activate_below_threshold:
            # Set point is below the threshold, turn output ON
            # e.g. Cooler
            output = set_point < self.threshold
        else:
            # Set point is above the threshold, turn output ON
            # e.g. Heater, Water leveler
            output = set_point > self.threshold

        return output

if __name__ == '__main__':
    OnOffOpen.start()
```

A couple of extra things:

 1. Should the names of the controllers be normalized into a common notation, e.g. always specifying if its _open or _closed control?
 2. Is it planned for boolean values to be allowed in recipes?
 3. If so, what value should be published to `/measured` as a fake measurement when being used with open-loop controllers. e.g. false: 0.0, true: 1.0? or should the sensor_persistence module also support std_msgs/Bool?

rbaynes:
No longer relevant.

PR #131, opened 2017-02-16
Replace openag_brain.var_types "vendored" var types with environmental variable category concept

gordonbrander:
This pull request removes the `openag_brain.var_types` file which we were using to associate var types with a "category" of variable. Instead of curating a list of variables, we introduce the concept of "categories" to var types in the `openag_python` package. This lets us get lists of variables per category type, and we can get rid of the openag_brain.var_types file, which too easily gets out of sync with this file.

This PR requires a complementary PR in openag_python https://github.com/OpenAgInitiative/openag_python/pull/35.


gordonbrander:
Tested on Raspberry Pi with https://github.com/OpenAgInitiative/openag_python/pull/35.

PR #133, opened 2017-02-17
Update env var references to use groups instead of categories

gordonbrander:
This is a naming change only per request from @LeonChambers  in https://github.com/OpenAgInitiative/openag_python/pull/35#issuecomment-280744509. It requires a corresponding PR in openag_python to land https://github.com/OpenAgInitiative/openag_python/pull/36.


gordonbrander:
@LeonChambers I've tested this on the Pi, and since I know it's just a naming change, I'm going to go ahead and merge it.

PR #137, opened 2017-02-22
Use roslaunch files instead of database for configuration

gordonbrander:
This PR offers an alternative configuration approach that hews closer to the standard ROS way of doing things:

- All nodes are configured with a `.launch` config file http://wiki.ros.org/roslaunch/XML
- All configuration parameters are loaded from yaml files into the [parameter server](http://wiki.ros.org/rospy/Overview/Parameter%20Server)

Before: `fixture -> load_fixture -> db -> launch file -> load config from db`
After: `launch file -> read config from param server`

[**Requires corresponding PR in openag_python**](https://github.com/OpenAgInitiative/openag_python/pull/37)

Rational
--------

These changes are the result of testing and working around problems experienced while setting up our first batch of FC2.0 machines.

This approach moves the system toward a largely static config system. Previously we would read changes made to the DB at runtime and attempt to restart ROS with the new config automatically. The rational was to allow for drag-and-drop reconfiguration of hardware at some future point. However, in practice this reconfigure-at-run-time layer caused problems (e.g. reflashing Arduino at inopportune times). This approach is simpler, but does not preclude fanciness at some future date, and in the meantime editing an XML or yaml config file is simple and robust.

The previous approach of `fixture file -> db -> generate launch file, connect topics, flash arduino -> launch` caused a number of practical footguns:

- Database state was cumulative... loading in fixture A, then fixture B would result in some merge of fixtures a and b. This made it hard to reproduce config states reliably when upgrading software.
- Changing firmware definitions was problem-prone because it required changing the fixture, changing the module file, loading the fixture and building a docker image with the updated fixture file. If any of these places didn't match, you would create runtime errors.

Additionally, it had these architectural side-effects:

- The fixture -> db -> launch system was complex
- Every rosnode ended up pretty tightly coupled with DB 

Changes
---------

After this PR, all system configuration happens through [roslaunch xml files](http://wiki.ros.org/roslaunch/XML). These files contain node definitions and may also load params into the param server via `.yaml` files (the ROS standard way of loading params).

We bundle several config files with the standard release:

- `personal_food_computer_v2.launch`
   - `personal_food_computer_v2.yaml` parameters describing firmware setup. This is used for codegen/flashing the arduino. The params in the parameter server are also used at runtime to connect the topics generated by the Arduino.
   - `usb_cam.yaml` default USB cam configuration

The Arduino flashing workflow is assumed to be a bit different. Rather than autoflashing (which caused problems in our tests), the Arduino is flashed once with a known good state, then left alone. This will also be better for release management. We can create a known good Arduino image and distribute it for flashing.

There are 2 supported workflows for flashing the Arduino:

- From a laptop, with [openag_python](http://github.com/openaginitiative/openag_python)
- From the RPi (usually over ssh), with [openag_python](http://github.com/openaginitiative/openag_python) installed in the global space.


gordonbrander:
See also: corresponding changes in https://github.com/OpenAgInitiative/openag_python/pull/37.

wsnook:
@gordonbrander I just saw your [RFC post](http://forum.openag.media.mit.edu/t/rfc-use-roslaunch-files-for-pfc-configuration/1502) on the form, so here's my two cents:

1. Your description of the changes you want to make sounds great. I've been studying the PFC2 design for a month now--my impression is that it fits the pattern of Fred Brooks's [second system](https://en.wikipedia.org/wiki/The_Mythical_Man-Month#The_second-system_effect) effect. Cutting features to [reduce the complexity](https://blog.codinghorror.com/kiss-and-yagni/) sounds good.

2. I have growing reservations about encouraging people to build the PFC2 design. On the positive side, it's a cool photogenic demo with great potential for introducing people to the idea of indoor farming--that's good. But, the design is expensive, complicated, and not practical for growing food. From the forums, it seems like people get excited about it, try to start building, then get stalled out. Very few people seem to be actually growing food or generating useful data. I think this is due in part to an emphasis on overly complicated technologies like ROS. For more on this line of thinking, see the "[How to move forward on sharing data?](http://forum.openag.media.mit.edu/t/how-to-move-forward-on-sharing-data/1485)" forum post.

3. As a counterpoint to #2 above, I'd like to see lots of CC BY-SA time lapse images of growing plants, including annotations on growing method, temperature, etc. If we're gonna run complicated software, let's make it something like using OpenCV to measure phenotype expression. I'm starting to play with these ideas over at https://github.com/wsnook/simplehydro. Pointing cameras at Kratky method lettuce grows seems like it might be ultimately more useful than a bunch of people fussing with ROS trying to get it do to something useful.

gordonbrander:
@wsnook thanks! I like all this feedback, but let's keep the conversation on this PR focused on code review. Can we follow up on your feedback 1:1 or in the forum?

wsnook:
@gordonbrander The reason I put these broader comments here is that you asked for comments on the forums. Having invested a lot of time going over the code and designs in recent weeks, these are the issues I care about now. Your proposed changes seem like a step in the right direction.

I'll DM you with my email.

gordonbrander:
A summary of the conversation so far:

- Favorable feedback on the PR because it simplifies the configuration stack.
- Others have pointed out that having hardware configuration information in the cloud may have some advantages in the long term. (This cloud doesn't yet exist, we want it to exist at some point)
- However, this PR solves pressing workflow problems with the current software and does not preclude the use of the DB for configuration at a future date, if that is deemed to be useful.

rbaynes:
As a n00b to ROS and this project, my input should be somewhat discounted.  But I completely agree with @gordonbrander 's rational.  This will let us move toward a system that is easier to deploy for non-engineering users.

LeonChambers:
Just gonna drop some comments here since I designed most of the original system.

So I think that at this stage of development, since we're looking to launch something soon, these changes makes sense. That being said, I'd like to clarify here some of the tradeoffs of making these changes.

There are several nice things about storing the module configuration in the database instead of in static files, but basically none of them were really being used yet.
- If we actually were syncing all of people's data to a cloud database and running analytics on it, it would be useful to be able to see exactly what software they are running in order to be able to correctly interpret their data.
- If we actually had a global database of module types, we could just sync those definitions down and would basically never have to use fixtures except for initial configuration based on the hardware a person has, which is a lot easier to reason about.
- At some point, it would be nice to have a UI for configuring firmware and software modules which would just build on top of the REST interface for CouchDB. This UI would probably be simpler than SSHing into the machine and editing config files.

Again, none of these things are implemented, so at the moment, storing module configuration in the database just serves as an additional level of indirection, making it harder to understand the system. Also, moving the configuration stuff from the config files back to the database isn't actually that hard if we want to implement some of these things later.

rbaynes:
Thank you @LeonChambers.  All great points about the design.  I will add them to the system notes I'm keeping in the wiki so we can think about adding those features back in once we have the back end database setup.

Issue #134, opened 2017-02-20
Ebb/Flow On/Off cycle controller?


Spaghet:
In aquaponics there is a need for something like a "5 minutes on, 10 minutes off" control for the circulation pump in order to conduct ebb/flow mechanics. Similar control should be useful in aeroponics as well.

Currently, one way to do this is the simple recipe doing 
[0, pump_var, 1], [300, pump_var, 0] repetitively, I think.
Is there an existing mechanic for this kind of control, or should I implement a software module to do this maybe?

LeonChambers:
You should just implement a software module to do this

Spaghet:
There seems to be a firmware module that does this called pulse_actuator but I haven't tried using it yet

PR #138, opened 2017-02-24
Implement ebb_flow_controller

Spaghet:
Fix to #134.

I'm kind of against defining the ebb/flow cycle in the fixture but I would need to end up changing a lot more things in order to implement something editable via recipe and I'm not entirely sure what the best practice for that is.

I've tested this to the degree that I see all the relevant rostopics(/environment/... and /actuator/...) oscillating between on/off, 1.0/0.0. Not sure if my hardware is doing it's job since I'm not next to it right now but in that case it is probably more of a hardware/arduino problem. 

image

### DIFF
--- a/src/openag_brain/software_modules/ebb_flow_controller.py
+++ b/src/openag_brain/software_modules/ebb_flow_controller.py
@@ -20,7 +20,7 @@ if __name__ == '__main__':
     on_time = rospy.get_param("~on_time", None)
     off_time = rospy.get_param("~off_time", None)
 
-    if on_time and off_time is None:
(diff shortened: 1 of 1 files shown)


Spaghet:
This seems to work for 12 hours and then crashes so I'm going to look into this a bit further

Spaghet:
This works now :)

rbaynes:
Has it run longer than 12 hours with your most recent fix?

Spaghet:
As of right now, it's been running for about 17 hours and ctrl+c/SIGINT works. I think it's safe to say it works and the bugs are flattened out.

PR #148, opened 2017-03-11
Topic filter docstring edits and test function for EWMA

rwdavis513:
Added docstrings to run through initial workflow


test/test_software_modules/test_topic_filer.py
```diff
@@ -0,0 +1,19 @@
+from openag_brain.software_modules.topic_filter import EWMA
```
gordonbrander: the name of this file should be `test_topic_filter.py`

src/openag_brain/software_modules/topic_filter.py
```diff
@@ -1,55 +1,77 @@
-#!/usr/bin/python
```
gordonbrander: on the diff of this file, every line has changed. I think this is because your editor is inserting Windows line endings. Please check to make sure your editor will insert Unix line endings https://help.github.com/articles/dealing-with-line-endings/.

test/test_software_modules/test_topic_filer.py
```diff
@@ -0,0 +1,19 @@
+from openag_brain.software_modules.topic_filter import EWMA
+import numpy as np
```
gordonbrander: Why are we importing numpy? Also, if we need it, we need to specify it as a dependency in `setup.py`.

test/test_software_modules/test_topic_filer.py
gordonbrander: Please also make sure that this file has Unix line endings.

PR #157, opened 2017-03-15
Update build setup to allow standalone distributions

gordonbrander:
This PR changes a few things to allow for the creation of standalone distributions:

- Move `src/openag_brain/software_modules` to `nodes`, per ROS convention
- Add macros to `CMakeLists.txt` to build `nodes`, `scripts`, `launch` and other supporting directories into standalone install.
- Add `openag_brain.commands` subpackage to `setup.py` to assure it is copied during install build.

## Motivation

Fixing our CMake build process will allow us to build standalone distributions. This means we can create a distribution, zip it up and add it to Github as a dist. Users will be able to install it on the Pi without having to build themselves. This also gives us the ability to do real releases, not just of the source, but of the distribution. We can also publish releases to `apt` package repo.

With this foundation, we should be able to create a trivial install script that simply `apt-get install`s CouchDB, ROS and openag_brain.

This is also an important step toward fixing our distribution approach for openag_ui and firmware. If we can produce stand-alone distributions, it means we can collect openag_ui and firmware into a monorepo.

## Usage

To create a distribution:

```bash
cd ~/catkin_ws
catkin_make install
```

A directory called `~/catkin_ws/install` will be created. Inside will be the full release of openag_brain, along with all of its dependencies and libraries. To run it, simply `source ~/catkin_ws/install/setup.bash` and then `rosrun openag_brain personal_food_computer_v2.launch`.

Additional background: http://wiki.openag.media.mit.edu/users/gordonb/build_tooling


CMakeLists.txt
```diff
@@ -152,6 +152,30 @@ catkin_package(
 #   DESTINATION ${CATKIN_PACKAGE_BIN_DESTINATION}
 # )
 
+# Mark executable scripts (Python etc.) for installation
+# in contrast to setup.py, you can choose the destination.
+#
+# We search for script files using the FILE(GLOB ...) macro, then assign the
+# result to openag_script_files.
+#
+# http://docs.ros.org/kinetic/api/catkin/html/howto/format2/installing_python.html
```
rbaynes: Excellent.  Thank you @gordonbrander for commenting with the sources you used to figure this out.

CMakeLists.txt
```diff
@@ -166,6 +190,15 @@ catkin_package(
 #   PATTERN ".svn" EXCLUDE
 # )
 
+# Install everything in the launch/ directory.
+# This CMake macro can also be used to install data files, etc.
+# See http://wiki.ros.org/catkin/CMakeLists.txt#Installing_roslaunch_Files_or_Other_Resources
+# - gordonb 2017-03-15
+install(DIRECTORY launch/
+  DESTINATION ${CATKIN_PACKAGE_SHARE_DESTINATION}/launch
+  PATTERN ".svn" EXCLUDE
```
rbaynes: @gordonbrander do we also want `PATTERN ".git" EXCLUDE`?

CMakeLists.txt
gordonbrander: I would love to make this a habit in our team. For any obtuse or dusty corner of the code, leave a note, any urls, a date and a name. That way if a dev truly gets stuck, they can reach out on GH, the forum or Slack to get unstuck.

CMakeLists.txt
gordonbrander: This is me being paranoid. ROS CMake formulas all include this EXCLUDE pattern. The reason is that when you use SVN for version control, it leaves `.svn` folders in every subfolder of the repo. Git does not do this, and manages its internal state in a top-level `.git` folder. I don't expect we will ever use SVN, but the exclude does no harm.

Thoughts?

rbaynes: Agree, just leave the check for .svn.   And forget about my suggestion for .git.

CMakeLists.txt
rbaynes: When we do PRs from anyone, we can ask that they do this before we will approve the change.   For something that needs more descriptions, diagrams, etc a URL to a document on the wiki will be good.

gordonbrander:
Also, because this PR moves around the repo layout, I made a tag for the state before it `0.1.1-alpha.2`

PR #159, opened 2017-03-29
(WIP, don't merge) Python peripherals

gordonbrander:
Based on work Jake did in https://github.com/OpenAgInitiative/openag_brain_box/, investigating how hard it would be to move all sensors and actuators to Python ROS nodes. This would mean the Arduino would no longer be necessary, reducing BOM cost and complexity.

Background notes: http://wiki.openag.media.mit.edu/contributors/gordonb/python_peripherals


wsnook:
@gordonbrander I saw the wiki page mentioned scripting raspi-config. If you're still looking for a way to script the raspi-config i2c stuff, try this:
```
sudo raspi-config nonint do_i2c 0
```

Here's a console log of me playing with `nonint` mode
```
$ # First see what it looks like with i2c turned off...
$ dmesg | grep i2c
[    2.344954] i2c /dev entries driver
$ lsmod | grep i2c
i2c_dev                 5859  0 
$ cat /boot/config.txt | grep i2c
dtparam=i2c_arm=off
$ 
$ # Now turn it on like this...
$ sudo raspi-config nonint do_i2c 0
$ 
$ # Check what changed...
$ dmesg | grep i2c
[    2.344954] i2c /dev entries driver
[  231.274035] bcm2708_i2c 3f804000.i2c: BSC1 Controller at 0x3f804000 (irq 83) (baudrate 100000)
$ cat /boot/config.txt | grep i2c
dtparam=i2c_arm=on
$ lsmod | grep i2c
i2c_bcm2708             4834  0 
i2c_dev                 5859  0
$  
$ # Now turn it back off...
$ sudo raspi-config nonint do_i2c 1
$ 
$ # and check what changed...
$ dmesg | grep i2c
[    2.344954] i2c /dev entries driver
[  231.274035] bcm2708_i2c 3f804000.i2c: BSC1 Controller at 0x3f804000 (irq 83) (baudrate 100000)
$ lsmod | grep i2c
i2c_bcm2708             4834  0 
i2c_dev                 5859  0 
$ cat /boot/config.txt | grep i2c
dtparam=i2c_arm=off
```

### Why this works
raspi-config is a big bash script that lives at `/usr/bin/raspi-config`. Basically, if you start with the `nonint` flag, it will interpret additional arguments as calls to functions with the script. If you search the script for `do_interface_menu()`, it should give you an idea of what's possible.

The `0` and `1` values correspond to an `if` block in `do_i2c()` that's checking the value of `$RET`. When the the `do_i2c` sees that `$INTERACTIVE` is true, it grabs the next argument after `do_i2c` with `RET=$1`, then that gets used to decide if i2c should be enabled or disabled.

serein7:
@gordonbrander I was able to run this POC on my environment with standalone RPI (i.e. no arduino attached) and see data from am2315 sensor published to air_temperature/raw topic. 

Now I guess I need just to add correct fixtures to make rest of the functionality working (i.e. saving to couchdb, filtering and etc). Is that correct?

gordonbrander:
@serein7 yeah, this PR is based on https://github.com/OpenAgInitiative/openag_brain/pull/158. Trying to simplify the stack and make it more reliable. Excited to hear you've got it working! I'm still hacking on it over here :D. I'm still prototyping this concept, but the idea would be to:

* Have simple nodes for reading sensor data
* Post sensor data to a topic named after the environmental variable.
* Remove Arduino from stack. Python is easier to hack on.
    * Eventually remove Arduino firmware stack.
* Remove `topic_filter`. Replace with filters in each Python module.
* Remove `topic_connector`. We won't need it anymore.

All in all, I suspect this setup would be cheaper, simpler and more reliable, since ROSserial (our connection to Arduino) has some limitations and bugs that have been tricky to work around.

serein7:
@gordonbrander, one thing I noticed after 24 hours of running this code with am2315 attached is that air temp and humidity values are weirdly jumpy http://joxi.ru/EA44anziWnZjAb 
I don't recall similar behaviour once it was connected to Arduino (though that was quite long time ago). Did you notice anything like this in your experiments?

gordonbrander:
This is because the sensors all occasionally make a bad read. topic_filter actually filters these jumps out from the Arduino sensors. If we have ROS nodes for each sensor, I think a better route to go would be to have the node itself filter noise out, rather than a separate node.

PR #161, opened 2017-04-07
Move var_types from openag_python to param server

rwdavis513:
Modified the structure of the var_types to be a yaml file which could be loaded directly into the ROS param server. The groups attribute for each variable was removed and changed into a nested dictionary, so all the recipe, environmental, camera and user variables are grouped within the same dictionary item (see yaml file for details).

### DIFF
--- a/launch/personal_food_computer_v2.launch
+++ b/launch/personal_food_computer_v2.launch
@@ -1,5 +1,6 @@
 <launch>
   <rosparam command="load" file="$(find openag_brain)/launch/personal_food_computer_v2.yaml" />
+  <rosparam command="load" file="$(find openag_brain)/launch/personal_food_computer_var_types.yaml" />
   <group ns="environments/environment_1">
     <node pkg="openag_brain" type="linear_controller.py" name="water_level_high_controller_1">
       <param name="variable" value="water_level_high" type="str"/>
--- a/launch/personal_food_computer_v2.launch
+++ b/launch/personal_food_computer_v2.launch
@@ -1,6 +1,6 @@
 <launch>
   <rosparam command="load" file="$(find openag_brain)/launch/personal_food_computer_v2.yaml" />
-  <rosparam command="load" file="$(find openag_brain)/launch/personal_food_computer_var_types.yaml" />
+  <rosparam command="load" file="$(find openag_brain)/launch/personal_food_computer_var_types.yaml" ns="var_types"/>
   <group ns="environments/environment_1">
     <node pkg="openag_brain" type="linear_controller.py" name="water_level_high_controller_1">
       <param name="variable" value="water_level_high" type="str"/>
--- a/launch/personal_food_computer_var_types.yaml
+++ b/launch/personal_food_computer_var_types.yaml
@@ -1,0 +1,82 @@
+recipe_variables:
+  recipe_start:
+    name: recipe_start
+    description: "Represents the start of a recipe"
+  recipe_end:
+    name: recipe_end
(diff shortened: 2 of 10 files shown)


src/openag_brain/load_env_var_types.py
```diff
@@ -0,0 +1,37 @@
+class VariableInfo(dict):
```
gordonbrander: Do we actually need this class, or can we just use the dict?

rwdavis513: Originally I set it up to use a dictionary. However, there are numerous places that it is referenced directly or as a key value lookup. For example,             
video_writer.py: 
    self.environment, "measured", variable.name,

I started going through and changing them all to use a dictionary, but later thought it would be more seamless of a change just to mimic the class structure so it could be used in the same way as it was previously.

thoughts?

gordonbrander: @rwdavis513 that makes sense. I appreciate that this lets us switch over without too many changes. We can always go back and remove the class later.

gordonbrander: Ok, will test.

launch/personal_food_computer_var_types.yaml
```diff
@@ -0,0 +1,82 @@
...
+    description: "A measure of the concentration of water in the air relative to the maximum concentration at the current temperature"
+    units: "percent relative"
+  air_carbon_dioxide:
+    name: air_carbon_dioxide
+    description: "The amount of Carbon Dioxide in the air"
+    units: "ppm"
+  air_flush_on:
+    name: air_flush_on
+    description: "Turn on air flush (off by default)"
+    units: "boolean"
```
gordonbrander: @rwdavis513 in this scenario we use "boolean" for units, but for our other boolean variable (water_level_high) we use no unit. We need to make this consistent.

Suggestion: we should have an additional field called `type` that uses the ROS type keys (`std_msgs/Float32`, etc) and use this to make decisions about how to filter channels.

rwdavis513: Great catch. I intended to at least have it consistent and use boolean for all of them, but looks like it slipped my mind in between working sessions. I seem to remember that most var types were Float64 already. Do you know if that's correct?

gordonbrander: @rwdavis513 you bring up a good point. All types are Float64 for all `environment` topics (booleans are encoded as 0 and 1). This is actually what caused the EWMA filter bug in the first place!

My intention is to have typed topics in future (std_msgs/Bool, std_msgs/Float64, etc), but I think that's too large of a change for this PR. Perhaps we can work around this by matching the var name(s) explicitly as we did before, then circle back in a follow-up PR to have true typed topics.

rwdavis513:
@gordonbrander I made the changes requested. Could you please take another look?

launch/personal_food_computer_v2.launch
```diff
@@ -1,5 +1,6 @@
 <launch>
   <rosparam command="load" file="$(find openag_brain)/launch/personal_food_computer_v2.yaml" />
+  <rosparam command="load" file="$(find openag_brain)/launch/personal_food_computer_var_types.yaml" />
```
gordonbrander: A small stylistic request: let's put all of these under an `environmental_variables` namespace. This can be done with the `ns` attribute. See http://wiki.ros.org/roslaunch/XML/rosparam. Will require changing the `get_param` calls in the other files.

PR #176, opened 2017-05-03
Include recipe_start and recipe_end in VALID_VARIABLES

gordonbrander:
Variables that aren't in the `VALID_VARIABLE` set are ignored. Since we missed including `RECIPE_START` and `RECIPE_END` in the `VALID_VARIABLE`s, it means we aren't storing a data point in the DB for those, meaning recipes will be unable to pick up across reboots of the Raspberry Pi.

Additionally, this proposes a fix for a (yet to manifest) bug caused by key ordering of dicts. Key order of dicts is not guaranteed by Python, but can be rather stable in a given implementation. We were counting on key order for `create_variables` to work properly.

Rather than reading a dict into a list, this PR lets the consumer decide how to read the dict.

A `from_dict` staticmethod is added to `VariableInfo` to give consumers an easier way to convert from YAML to class instance.

### DIFF
--- a/nodes/image_persistence.py
+++ b/nodes/image_persistence.py
@@ -96,7 +96,7 @@ if __name__ == '__main__':
         min_update_interval = 3600
     env_var_db = server[ENVIRONMENTAL_DATA_POINT]
     persistence_objs = []
-    for variable in CAMERA_VARIABLES:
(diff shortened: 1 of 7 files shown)


gordonbrander:
cc @rbaynes

Issue #130, opened 2017-02-14
Way to break feedback loops when recipe is finished/paused


gordonbrander:
Once loops are started, they never stop. We need a way to set "null" setpoints for loops so we can put machine in idle mode.

Spaghet:
This was fixed in #177, broken again, and then re-fixed in #311

PR #177, opened 2017-05-03
Null setpoints with sentinel values

gordonbrander:
This PR introduces the concept of "null setpoints", or the ability to break controller feedback loops. It is a fix for https://github.com/OpenAgInitiative/openag_brain/issues/130.

## Rationale

Recipes signal a desired setpoint. From there, it is the job of the controller to plan and continue sending values to the actuator to achieve that goal. Sometimes it's desirable to break a feedback loop — to put the controller back into a null state.

The primary motivating example is ending a recipe. At the end of a recipe, we want to be able to put everything back into an idle state (no heater, no cooler, no light, etc).

## Details

ROS does not support polymorphic or Union types for typed topics. A common workaround is to use [Sentinel values](https://en.wikipedia.org/wiki/Sentinel_value) ("magic numbers") to signal a specific desired change of state.

In this PR we use a sentinel Float64 value to mean "break this feedback loop", since all environmental variables are currently encoded as Float64.

## Alternatives

An alternative could be to have a separate ROS topic (for example ''idle'') in a given environment. In this scenario all controllers would listen to that topic in addition to their topic, and break the feedback loop if a message came through on it.

While this approach does not require a sentinel value, it means we can not break loops on an individual basis.

## Future

In future, we plan for environment topics to use ROS types appropriate to their data type, rather than encoding everything Float64, as we currently do.

When environments support more types than Float64, we should create new message struct types that have a ''data'' field (containing the payload) and a ''should_break_feedback: bool'' field. Controllers could then check the ''should_break_feedback'' field instead of checking for a sentinel value.


gordonbrander:
@rwdavis513 I think this PR is ready to go, but I'm having trouble testing it on the dev bot in Utah. I'm betting that bot has electronics issues.

Can you specify a bot that we can use to test this?

nodes/pid.py
```diff
@@ -41,7 +42,12 @@ def __init__(self, Kp=0, Ki=0, Kd=0, upper_limit=1, lower_limit=-1,
         self.integrator = 0
 
     def update(self, state):
+        # Set a null setpoint if we see the magic number.
+        if state is NULL_SETPOINT_SENTINEL:
+            self.set_point = None
+        # If setpoint was made null, or was already null, do nothing.
         if self.set_point is None:
+            self.set_point = None
```
rwdavis513: Is this line needed? If set_point is None, then why set it equal to None?

rwdavis513:
@gordonbrander I'm getting this error when trying to run this code. Could you take a look?

process[api-19]: started with pid [778]
Traceback (most recent call last):
  File "/home/pi/catkin_ws/src/openag_brain/nodes/direct_controller.py", line 18, in <module>
    from openag_brain.constants import NULL_SETPOINT_SENTINEL
  **File "/home/pi/catkin_ws/src/openag_brain/src/openag_brain/constants.py", line 18, in <module>
    NULL_SETPOINT_SENTINEL
TypeError: 'float' object is not iterable**
[environments/environment_1/nutrient_flora_duo_a_controller_1-12] process has died [pid 723, exit code 1, cmd /home/pi/catkin_ws/src/openag_brain/nodes/direct_controller.py __name:=nutrient_flora_duo_a_controller_1 __log:=/home/pi/.ros/log/9e5da5ec-3594-11e7-830d-b827ebff173d/environments-environment_1-nutrient_flora_duo_a_controller_1-12.log].
log file: /home/pi/.ros/log/9e5da5ec-3594-11e7-830d-b827ebff173d/environments-environment_1-nutrient_flora_duo_a_controller_1-12*.log
process[topic_connector-20]: started with pid [781]
[environments/environment_1/nutrient_flora_duo_b_controller_1-13] process has died [pid 724, exit code 1, cmd /home/pi/catkin_ws/src/openag_brain/nodes/direct_controller.py __name:=nutrient_flora_duo_b_controller_1 __log:=/home/pi/.ros/log/9e5da5ec-3594-11e7-830d-b827ebff173d/environments-environment_1-nutrient_flora_duo_b_controller_1-13.log].
log file: /home/pi/.ros/log/9e5da5ec-3594-11e7-830d-b827ebff173d/environments-environment_1-nutrient_flora_duo_b_controller_1-13*.log
process[sensor_info_publisher-21]: started with pid [786]
Traceback (most recent call last):
  File "/home/pi/catkin_ws/src/openag_brain/nodes/direct_controller.py", line 18, in <module>
    from openag_brain.constants import NULL_SETPOINT_SENTINEL
  File "/home/pi/catkin_ws/src/openag_brain/src/openag_brain/constants.py", line 18, in <module>
    NULL_SETPOINT_SENTINEL
TypeError: 'float' object is not iterable

gordonbrander:
@rwdavis513 taking a look

gordonbrander:
@rwdavis513 fixed. Was missing a trailing comma (making `()` a tuple).

rwdavis513:
@gordonbrander Just to make sure, we're on the same page, I don't see any function to trigger sending the sentinel values effectively pausing the recipe. To test it out I was going to write a recipe which included the sentinel value. Is that what you were thinking?

gordonbrander:
@rwdavis513 the code can be seen at the end of the recipe interpreters. Before yielding a `RECIPE_END`, they will send a `NULL_SETPOINT_SENTINEL` to each environmental variable topic https://github.com/OpenAgInitiative/openag_brain/pull/177/files?diff=unified#diff-5ca20bc1ad7bdc054a4e1f4fa9cebfcbR130

PR #179, opened 2017-05-11
Improve the reliability of the camera

Spaghet:
## Summary

Use the ROS `aerial_image/image_raw` topic instead of fswebcam for improved camera reliability.

## Motivation

The current camera footage relies on [fswebcam](https://www.sanslogic.co.uk/fswebcam/). A picture from the camera is saved to `/var/www/html/img.jpg` and served over nginx. fswebcam will attempt to take a picture once every 10~15 seconds or so, but will often fail and not retry. The UI will attempt to reload the image once every 26 seconds or so. 

## End Goal

Ideally, the end goal for this would be to serve the camera footage over mjpeg or something so that we can just use an `img` tag to serve the camera footage. In the interim, I want to land this PR in order to provide something more consistent. 

## Details

This installs a few more ROS packages with openag_brain, namely [`image_view`](http://wiki.ros.org/image_view). This relies on some other packages like `cv_bridge`.
I've added these dependencies to the install scripts.

`image_view` has a node called `image_saver` which takes image topics of message type `sensor_msgs/Image` and saves an image based on it. I've simply taken `image_saver` and pointed it to `/var/www/html` to reduce the amount of code changes since this is a temporary fix to the bottleneck and not meant as a long term fix.

## Alternatives

I looked into a different mjpg server package, [web_video_server](http://wiki.ros.org/web_video_server). 

## Known problems

1. The install will sometimes fail on `catkin_make_isolated` at `cv_bridge` but if you go into `catkin_ws/build_isolated/cv_bridge` and run `sudo make` it will pass without failure. 
1. The minimum framerate for the camera is 1 FPS, but that means that the image file can be rewritten while it is being transported over the network.

## Working notes
http://wiki.openag.media.mit.edu/contributors/rikuo/ros/usb_cam


launch/personal_food_computer_v2.launch
```diff
@@ -1,7 +1,9 @@
 <launch>
+  <arg name="environment_1" value="environments/environment_1" />
+
```
novemberalpha: remove blank line?

scripts/init_dev
```diff
@@ -38,10 +38,15 @@ sudo apt-get update -q
 # sudo apt-get -y upgrade
 
 # Install pip and couchdb (couchdb is not available via rosdep -- see below)
-echo "Install pip, couchdb and boostrapping tools"
-sudo apt-get install -y -q python-pip couchdb
+echo "Install pip, couchdb and bootstrapping tools"
+sudo apt-get install -y -q couchdb
+sudo easy_install pip
```
novemberalpha: I had to manually do this. python-pip is flakey. Thanks for adding this in!

gordonbrander: @novemberalpha can you clarify? Was `python-pip` flakey or was `sudo easy_install pip` flakey? How did it fail?

gordonbrander:
> The install will sometimes fail on catkin_make_isolated at cv_bridge but if you go into catkin_ws/build_isolated/cv_bridge and run sudo make it will pass without failure.

- Do we know why?
- Do we have a fix that works consistently?
- Does it fail on a `catkin_make install` (not isolated)?

scripts/init_dev
novemberalpha: @gordonbrander I was trying to re-run the install script after I suspected it had failed in a prior attempt (I ran prior headless). Rerunning it gave `ImportError: cannot import name IncompleteRead` error at the pip commands [[1](https://github.com/OpenAgInitiative/openag_brain/blob/master/scripts/init_dev#L44), [2](https://github.com/OpenAgInitiative/openag_brain/blob/master/scripts/init_dev#L86)] in the init_dev file (I commented one out and it failed on another). I found [this solution](http://stackoverflow.com/questions/27341064/how-do-i-fix-importerror-cannot-import-name-incompleteread) and was able to run the re-install.

gordonbrander:
@novemberalpha btw, thank you for testing and reviewing! Please keep doing this! We could use the help!

scripts/init_dev
Spaghet: Yeah using apt-get python-pip was flakey

Spaghet:
@gordonbrander No idea, it does work usually, seems to happen sometimes. I thought it was an opencv dependency thing but that doesn't seem to be the case. I was hoping to run `sudo make -d` to debug it and it just went through and passed. The builds taking an hour and very helpful to do iterations.

I've yet to try `catkin_make install`.

launch/personal_food_computer_v2.launch
Spaghet: That's there since args are like configurable parameters for the launchfile, and it's a fairly common launchfile thing to do. Think of them like a file-wide CONST_VARIABLE.

gordonbrander: @Spaghet that's pretty cool

scripts/init_dev
gordonbrander: @rwdavis513 would this also solve the `Requests` versioning conflict we have with `apt-get`'s version of `pip`?

Spaghet: That is the problem, `requests` deprecated the `IncompleteRead` API and ubuntu/apt-get didn't catch up.

See more here:
http://stackoverflow.com/questions/27341064/how-do-i-fix-importerror-cannot-import-name-incompleteread

gordonbrander: @Spaghet thanks! We should land this in a separate PR. I can make it if you want.

gordonbrander: @Spaghet https://github.com/OpenAgInitiative/openag_brain/pull/180

PR #255, opened 2017-05-22
[WIP] Stateless Recipe Interpreter

gordonbrander:
Work-in-progress PR for #254

PR #257, opened 2017-05-23
Decouple recipe persistence

Spaghet:
WIP

Related to #255, just takes the `environment_variable/desired`  and saves it to the couchdb environment_variable database.

It was a @TODO under the recipe_handler node to decouple so I figured I would just do that


launch/personal_food_computer_v2.launch
```diff
@@ -29,6 +29,7 @@
       <param name="Kp" value="1" type="double"/>
     </node>
     <node pkg="openag_brain" type="recipe_handler.py" name="recipe_handler_1"/>
+    <node pkg="openag_brain" type="recipe_persistence" name="recipe_persistence_1"/>
```
rbaynes: Should be `type="recipe_persistence.py"` add .py ?

Spaghet:
Yeah, I'm trying to test it right now but it seems to run fine after we add the .py extension

rwdavis513:
@Spaghet Could you add a test node to verify this is working correctly? I'd like to get in the habit of including tests for all new code that is submitted.

Spaghet:
Do we have a specific testing tool we want to adhere to?
I see at least 3 we're using... nosetest, assert, unittest.

I'm going to get the new recipe interpreter running, and work with rbaynes on how ROStest works and get to testing this before I merge it in.

PR #259, opened 2017-05-24
Move Ebb and Flow controller to nodes/ directory

Spaghet:
The other software_modules got moved to nodes/ but ebb_flow_controller got left behind.


Spaghet:
@rbaynes @rwdavis513  I just moved a file, if someone could merge this that would be nice, I'm going to be using this node soon

Issue #254, opened 2017-05-22
[Exploration] Stateless Recipe Interpreter


gordonbrander:
- Concept: create a stateless recipe interpreter
- Working branch: https://github.com/OpenAgInitiative/openag_brain/tree/recipe_stateless

## Rationale

The current recipe interpreter is rather complex... probably the most complex and difficult-to-test portion of code outside of the interface with rosserial.

- Currently, recipe interpreters do not allow you to "rewind"... because it is a stateful iterator it can only seek forward.
- Because the recipe interpreter uses a blocking timer and is stateful, it is very hard to test.
- `recipe_handler.py` uses a combination of sugars and unusual features that are hard to grok (generators, custom iterators, decorators, blocking timers, threading and locking, memoization, multimethods). Mia culpa.

If we make the interpreter a stateless function, it would be easy to test, and we could use the [standard ROS event loop pattern](https://github.com/OpenAgInitiative/ros_examples/blob/master/minimal_ros_node.py) for timing.

## Details

Approach:

- Recipe interpreter should be a stateless function
  - Unit test interpreter functions
- Put the code that stores environmental data points in DB in a separate ROS node
  - This is in keeping with single responsibility principle
  - It also means we can turn on/off the database storage by including/removing from launch file

`RecipeHandler` becomes much simpler. It is just a threadsafe way to set and retrieve the document for the currently running recipe. It is responsible for:

- Keeping the current recipe document
- Keeping the current recipe start time
- Fetching recipes from DB
- Providing threadsafe methods for `set`ing, `clear`ing and `get`ing recipe.
- Registering ROS services

Note that methods should be threadsafe because ROS services run in a separate thread.

cc @rbaynes @Spaghet

Spaghet:
I like it, allows for recipe scrubbing as mentioned in #79  much easier.

wsnook:
@gordonbrander I don't understand what you mean by a stateless interpreter. As I understand interpreters, maintaining state is an essential part of what they do (e.g. call stacks, variables, current instruction being executed, etc.).

Do you mean you want to write a compiler that translates recipe source code into an intermediate representation--maybe a list of scheduled actions for changing control parameters? If so, that's a great idea. You could unit test the scheduler, unit test the modules that implement the configurable control loops, and unit test that recipe source files produce the correct compiled output.

gordonbrander:
[...]

PR #263, opened 2017-05-30
Stateless recipe interpreter

rwdavis513:
Built on top of [PR255](https://github.com/OpenAgInitiative/openag_brain/pull/255) to address issue [#254](https://github.com/OpenAgInitiative/openag_brain/issues/254).

Added tests to verify recipe interpreter was working on a small mock recipe.

Fixed test_load_env_var_types so that it is now a passing test.


rbaynes:
@rwdavis513 I ran the test for the recipe_handler like so:
`rostest openag_brain test_recipe_handler.launch`
and it looks like it succeeds:
```SUMMARY
 * RESULT: SUCCESS
 * TESTS: 1
 * ERRORS: 0
 * FAILURES: 0
```
but when I look at `~/.ros/log/latest/environments-test-test_recipe_handler-2.log` I see this error:
```
 Unable to report rosout: publish() to a closed topic
```
but from looking at the other logs, I guess rosout is not started for testing, so this is expected and doesn't affect the test.

Issue #194, opened 2017-05-12
Create new Recipe_handler to process new time format. Start/stop recipes (3d)


gordonbrander:
https://github.com/OpenAgInitiative/openag_brain/issues/79

- Pause means &quot;break all feedback loops&quot;.
  - Depends on https://trello.com/c/H0WGwXJA/912-way-to-break-feedback-loops-when-recipe-is-finished-paused-null-setpoints-2d

(Exported from Trello https://trello.com/c/EPANisO4/564-create-new-recipe-handler-to-process-new-time-format-start-stop-recipes-3d)

---

On 2017-04-27T17:55:52.387Z, @bobdavis27 wrote:

 (add pause, seek, start @ given time) 

---

On 2017-03-30T20:37:46.085Z, @gordonbrander wrote:

This is required for https://trello.com/c/b5i4IQdy/516-update-ui-control-panel-to-retain-state-override-recipe.

rwdavis513:
This was resolved with PR#266.

PR #266, opened 2017-06-02
New recipe format called flexformat + more tests

rwdavis513:
This new recipe format allows for any complexity that the simple recipe format can provide, but with the smaller file size of phased recipes. 

Recipe Format structure:
phase:
    step: { 
         variable: [start_time, end_time, value] 
    }

The phases can be repeated as many times as needed as well as the steps. Each step is ideally a 24 hour period.

See `openag_brain/data/mock_recipes` for an example

This new format is compatible with the cloud software.


rwdavis513:
Note: The CouchDB setup on the PFC needs to allow for the recipes database to have a phases field instead of operations (or we could update the new format to use operations instead of phases).

`couchdb.http.ServerError: (403, (u'forbidden', u'Recipe instances are required to have a operations field'))`

rbaynes:
Some suggestions:

1. If we use "operations" instead of "phases" in the JSON of a recipe, we avoid the DB change.
1. There are existing recipes and users currently using a recipe with the format name of "phased".  I think we should pick a new name for this new format to avoid confusion.  Perhaps call this new format "detailed" or anything you like.
1. Thanks for breaking things up into their own files.
1. I intended TRACE/trace() to be per node, but rospy.logdebug() does that already, so it will work.
1. After reviewing my changes to the tests (PR 267) and merging to master, can we merge those changes into this branch?

rwdavis513:
@rbaynes I merged your changes from [PR267](https://github.com/OpenAgInitiative/openag_brain/pull/267) this PR. I agree a different name would be better. I started with "phased_dense" as it is a phased recipe that can contain a dense amount of information. I'd be glad to change it to anything else though. Unfortunately, I used the name quite a few places so I'll need to go back through to update the name to something else.

rwdavis513:
@rbaynes It looks like it would take the cloud team about a day to change from phases to operations, so I'll look into updating the couchdb to handle both for now or remove the restriction.

rwdavis513:
@rbaynes I updated the recipe name to "flexformat" to create a bigger difference from the prior recipe formats. Rather than changing "phases" to "operations" I'd like to remove the "operations" requirement check from the couchdb doc database. This is a simple one line change.  See [openag_python PR#41](https://github.com/OpenAgInitiative/openag_python/pull/41)

rwdavis513:
This should be ready to be merged into master. Let me know if you see anything else. Thanks!

rwdavis513:
Addresses Issue [#194](https://github.com/OpenAgInitiative/openag_brain/issues/194)

rbaynes:
Love the "flexformat"  💪

Spaghet:
Is there a reason there are variable units in the recipe? We already have that information in the var_types param server which will allow us to maintain a single reference point to what we mean when we say "light_illuminance" or whatnot. We have a problem where we declare the same thing in multiple places and they all need to sync up.

PR #271, opened 2017-06-07
[WIP] Bespoke serial

Spaghet:
There's a couple TODOs left in this one once I push the new `arduino_handler` node up.

We need to measure the pump rate when it's at maximum, to set the appropriate constant in the `openag_doser_pump` firmware module class.

I also talked to @jakerye and I'm going to restructure the way we're handling `air_flush` a bit.

Edit: Here are some notes on what I am doing here. https://wiki.openag.media.mit.edu/contributors/rikuo/serial


Spaghet:
This change also breaks the UI toggles for the hardware since we deprecated the /sensor and /actuator topics. They had no business being there in the first place, but alas.

rbaynes:
I'm taking care of killing the stuff in openag_python/_design and utils that you don't need anymore:
* safe_cpp_var()
* DB Firmware* and Software*

Spaghet:
Nice! I've deleted them (without commiting those changes) and it seems to cause no trouble for what I'm working on.

Spaghet:
Ok, I'm done with this one. Time to get into testing :)

gordonbrander:
🎉🏅🏆

Spaghet:
I think this is ready to get merged in, but we're going to be doing this into `develop` iirc so I'll close this and make a new one.

PR #273, opened 2017-06-09
Modify time format for flexformat recipe interpreter

rwdavis513:
The variable type for start_time and now_time did not match the simple recipe interpreter, which is now fixed. 

Also, added more tests to verify functionality.

### DIFF
--- a/nodes/recipe_handler.py
+++ b/nodes/recipe_handler.py
@@ -246,6 +246,7 @@ class RecipeHandler:
 #------------------------------------------------------------------------------
 # Our ROS node main entry point.  Starts up the node and then waits forever.
 if __name__ == '__main__':
+    pub_debug = rospy.Publisher('debug/recipe_handler', String, queue_size=10)
     if TRACE:
         rospy.init_node("recipe_handler", log_level=rospy.DEBUG)
     else:
@@ -289,6 +290,8 @@ if __name__ == '__main__':
 
             # Get recipe state and publish it
             setpoints = interpret_recipe(recipe_doc, start_time, now_time)
+            rospy.loginfo("Start: {} now_time: {}: ".format(start_time, now_time))
+            print(start_time, now_time)
             for variable, value in setpoints:
                 try:
                     pub = PUBLISHERS[variable]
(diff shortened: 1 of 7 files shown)


nodes/recipe_handler.py
```diff
@@ -246,6 +246,7 @@ def save_recipe_dp(self, variable):
 #------------------------------------------------------------------------------
 # Our ROS node main entry point.  Starts up the node and then waits forever.
 if __name__ == '__main__':
+    pub_debug = rospy.Publisher('debug/recipe_handler', String, queue_size=10)
```
rbaynes: Do we want to make `pub_debug` conditional on `TRACE`?

src/openag_brain/recipe_interpreters.py
```diff
@@ -58,20 +68,39 @@ def interpret_simple_recipe(recipe, start_time, now_time):
...
-    Recipe Interpreter should read a recipe, current_time, start_time, variable and return a value.
+    Recipe Interpreter should read a recipe, now_time, start_time, variable and return a value.
        Determine the time since the beginning of the recipe.
        Determine what is the current step
        Calculate the remaining time left in this step.
        Look up the value within that step for that variable.
     """
-
+    _id = recipe["_id"]
+    rospy.loginfo(recipe["phases"])
```
rbaynes: should be logdebug() ?

src/openag_brain/recipe_interpreters.py
```diff
@@ -83,6 +112,28 @@ def interpret_flexformat_recipe(recipe, current_time, start_time):
     )
 
 
+def verify_time_units(time_var):
```
rbaynes: 👍

src/openag_brain/utils.py
```diff
@@ -4,6 +4,7 @@
 from random import randint
 from re import match
 import json
+import rospy
```
rbaynes: Do we need this import?

src/openag_brain/utils.py
```diff
@@ -45,7 +46,7 @@ def read_environment_from_ns(namespace):
 # Turn on logic tracing by making the variable below True.
 # Output ONLY is written to this node's log file:
 # tail -f ~/.ros/log/latest/environments-environment_1-recipe_handler_1-6.log
-TRACE = False   # TODO: Move to settings file?
+TRACE = True   # TODO: Move to settings file?
```
rbaynes: Should we check this in as False?

rwdavis513:
@rbaynes Thanks for catching those items. I've updated the code and pushed my latest commits. Could you take another look?

PR #274, opened 2017-06-12
Bespoke serial protocol

Spaghet:
## Summary
This change replaces rosserial with a bespoke serial communications protocol. The goal of this was to create a faster, more reliable connection between the Arduino and the Raspberry Pi. This change is also a step towards using individual nodes for each hardware component, so we can use USB sensors and start writing our devices in python. 

## Motivation
rosserial was notoriously flaky, so we wanted to either fix it or replace it. We decided to replace it, and work on a new serial communication protocol that would be more reliable.
We also wanted to remove the topic_connector node since we could just have the sensors publish to `/environments/environment_1/environment_variable/raw` instead of mapping /sensors/ over to the same thing. 
This is mainly an interim fix for the RPi Arduino combination we currently run, and we will hopefully move on to something like BeagleBone or Standalone RPi in the near future.

## Details
First, all the firmware_module classes are now non-blocking and no longer contain ROS code. They also return `status_level` s (currently OK, WARN, and ERROR) which allow the main Arduino sketch to figure out if there is an error or not.

Second, the serial protocol is implemented in such a way that the "state" of sensors and actuators are passed around. The commands are not "Turn on for 15 seconds", but "It should be on". The firmware modules have a safety mechanism to shutoff after a set amount of time, usually 10 sec., so it is not going to be disastrous if the ROS side fails. The serial loop is running at around 50 Hz, and the ROS loop is configured to be around 1Hz. 

The topics that the arduino_handler node listens on is also hard_coded, as mapping between actuator and environment_variable is not declared in an efficient way for this implementation. The publishing of sensor values happens in a way that is less hard coded as it was mapped more directly. 

## Drawbacks
This change removes the codegen  and dynamic configuration capabilities from the firmware side of things.
It also hard-codes the data types for the Arduino sensors and actuators without referencing the var_types file or the other params since it was meant to be tightly coupled with the Arduino and not all variables map in  a simple way.

## Alternatives

 - rosserial
 - BeagleBone
 - standalone sensors, standalone nodes

Working notes: https://wiki.openag.media.mit.edu/contributors/rikuo/serial


rbaynes:
Wow, that is the best PR description evah!  Reviewing now.

PR #275, opened 2017-06-13
Deprecate openag_python, move remaining functionality to openag_brain/.../openag_lib module.

rbaynes:
When reviewing, please run the tests!   
`py.test ~/catkin_ws/src/openag_brain/test/test_lib/`


scripts/init_dev
```diff
@@ -47,6 +47,9 @@ sudo apt-get install -y -q couchdb python-setuptools
 sudo easy_install pip
 sudo pip install rosdep rosinstall_generator wstool rosinstall
 
+# Install some python dependencies that there aren't ros packages for.
+sudo pip install voluptuous HTTPretty pytest
```
Spaghet: Do we use HTTPretty for testing or something?

rbaynes: @Spaghet yes, in the tests, along with pytest.  It is a cool module.

Issue #166, opened 2017-04-24
event loop crashes when pi loses sync with arduino and port changes


jakerye:
## Summary
When running the recipe, the arduino will periodically lose sync with the raspberry pi. When this happens, the pi restarts the arduino. When the arduino restarts, sometimes the port will change from ttyACM0 to ttyACM1. This breaks the event loop on the raspi.

## Docker logs
```
[INFO] [WallTime: 1492639648.890978] Setup publisher on /sensors/water_level_sensor_high_1/is_on/raw [std_msgs/Bool]
[INFO] [WallTime: 1492639648.912157] Setup publisher on /sensors/atlas_ph_1/water_potential_hydrogen/raw [std_msgs/Float32]
[INFO] [WallTime: 1492639648.932244] Setup publisher on /sensors/atlas_ec_1/water_electrical_conductivity/raw [std_msgs/Float32]
Traceback (most recent call last):
  File "/home/pi/catkin_ws/src/rosserial/rosserial_python/nodes/serial_node.py", line 85, in <module>
    client.run()
  File "/home/pi/catkin_ws/src/rosserial/rosserial_python/src/rosserial_python/SerialClient.py", line 503, in run
    self.requestTopics()
  File "/home/pi/catkin_ws/src/rosserial/rosserial_python/src/rosserial_python/SerialClient.py", line 389, in requestTopics
    self.port.flushInput()
  File "/usr/lib/python2.7/dist-packages/serial/serialposix.py", line 500, in flushInput
    termios.tcflush(self.fd, TERMIOS.TCIFLUSH)
termios.error: (5, 'Input/output error')
```
## Similar issues: *rosserial-running-please*
Source
- http://answers.ros.org/question/233355/rosserial-running-please/

Make sure when a new device is connected, a new file is created. Check which port the arduio is connected to:
```
ls -la /dev/ttyACM*
```

Output:
```
crw-rw---- 1 root dialout 166, 1 Apr 19 16:40 /dev/ttyACM1
```

Hypothesis: 
- The arduino first connected as ttyACM0, when device lost sync and restarted, the arduino reconnected as /dev/ttyACM1?

Test:
- Unplug arduino, replug arduino, check port arduino is connected to again

Results
- Arduino now connected as ttyACM0

Next steps:
- Figure out a better method for handling when the arduino loses sync with the raspi.
- Figure out why arduino loses sync with raspi..but that is a separate issues

gordonbrander:
[...]

PR #279, opened 2017-06-15
More reliable arduino serial port assignment

Spaghet:
This is a fix to #166 regarding the Arduino.

Serial devices get mapped under `/dev/serial/by-id/` with a long and complicated device unique ID, but arduino devices include the phrase "arduino" in this ID.
We detect this and use that serial device instead of the `/dev/ttyACM0` since that name can sometimes change.


Spaghet:
oh woops I did this to master

PR #284, opened 2017-06-16
Minor fixups while testing develop branch

rbaynes:
Still a bit of serial work to do.  Working with Rikuo @Spaghet


scripts/deprecate_openag_python_and_upgrade_dev_install
```diff
@@ -12,5 +12,7 @@ set -e
 sudo mv /usr/local/bin/openag /usr/local/bin/openag.deprecated
```
Spaghet: I would rather this be a gist or something and not checked into the repo. We should take it out if we want to merge this to master. As for develop I'm ok with it being in if we remove it soon.

rbaynes: agree.  it is really just there for the short term

PR #289, opened 2017-06-22
Added some handshake logic to the serial protocol

rbaynes:
Now there is a 1:1 message transfer with ROS-arduino_handler.py as the master/initiator of the communication.  The Arduino sketch waits until it receives a message before it responds.

### DIFF
--- a/firmware/src/src.ino
+++ b/firmware/src/src.ino
@@ -42,6 +42,7 @@ ToneActuator chiller_compressor_1(9, false, 140, -1);
 // Message string
 String message = "";
 bool stringComplete = false;
+bool receivedFirstMessage = false;
 const int COMMAND_LENGTH = 18; // status + num_actuators
 const unsigned int MESSAGE_LENGTH = 500;
 
@@ -131,6 +132,7 @@ void serialEvent() {
     // so the main loop can do something about it:
     if (inChar == '\n') {
       stringComplete = true;
+      receivedFirstMessage = true; // first contact! (handshaking)
       //send_warning("debug Arduino serialEvent complete", message);
       return;
     }
@@ -190,6 +192,9 @@ void actuatorLoop(){
 //-----------------------------------------------------------------------------
 // Run the update loop
 void updateLoop(){
+  if(! receivedFirstMessage){ // don't send any serial data until first contact
+    return;
+  }
   bool allActuatorSuccess = true;
 
   allActuatorSuccess = updateModule(pump_1_nutrient_a_1, "Pump 1 Nutrient A") && allActuatorSuccess;
--- a/firmware/src/src.ino
+++ b/firmware/src/src.ino
@@ -45,15 +45,23 @@ bool stringComplete = false;
 const int COMMAND_LENGTH = 18; // status + num_actuators
 const unsigned int MESSAGE_LENGTH = 500;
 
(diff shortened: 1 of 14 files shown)


firmware/src/src.ino
```diff
@@ -118,53 +132,70 @@ void serialEvent() {
...
 // #endregion
 
+
+//-----------------------------------------------------------------------------
 void actuatorLoop(){
   // If serial message, actuate based on it.
-  if(stringComplete){
-    String splitMessages[COMMAND_LENGTH];
-    for(int i = 0; i < COMMAND_LENGTH; i++){
-      splitMessages[0] = "";
```
rbaynes: fun bug here!   `splitMessages[0]` is changed to `[i]` farther down in the diff.

firmware/src/src.ino
```diff
@@ -222,11 +250,37 @@ void sensorLoop(){
...
+  }
+  message = "";
+  stringComplete = false;
+}
+
+//-----------------------------------------------------------------------------
 int split(String messages, String* splitMessages,  char delimiter){
   int indexOfComma = 0;
   int delim_count = 0;
-  for(int i = 0; messages.indexOf(delimiter, indexOfComma) > 0; i++){
```
rbaynes: If you get a too long message, you will write out of bounds in splitMessages and stomp on memory!  So only read up to a maximum limit (of the size of splitMessages).

nodes/arduino_handler.py
```diff
@@ -273,38 +275,43 @@ def process_message(line):
...
             for headers, value in zip(sensor_csv_headers[1:], variable_values))
         return pairs
     except ValueError:
-        message = "Type conversion error, skipping."
+        message = "arduino_handler: Type conversion error, skipping."
         rospy.logwarn(message)
         return message
     except IndexError:
-        message = "Short read, received part of a message: {}".format(buf.decode())
+        message = "arduino_handler: Partial message: >{}<".format(line)
```
Spaghet: This will make the test fail since the message contents changed, btw.

nodes/arduino_handler.py
```diff
@@ -273,38 +275,43 @@ def process_message(line):
...
     except UnicodeDecodeError:
-        message = "Received weird bits, ignoring: {}".format(buf)
+        message = "arduino_handler: Ignoring weird bits: >{}<".format(line)
         rospy.logwarn(message)
         serial_connection.close()
         serial_connection.open()
         return message
 
+
+#------------------------------------------------------------------------------
```
Spaghet: We don't need these artifacts

rbaynes: But, I like those 'artifacts'!  It makes it easier for me to read the code since it doesn't have {} blocks like a real language...     I will remove them if they offend you.

src/openag_brain/settings.py
```diff
@@ -3,7 +3,9 @@
 # Turn on logic tracing by making the variable below True.
 # Output ONLY is written to this node's log file:
 # tail -f ~/.ros/log/latest/environments-environment_1-recipe_handler_1-6.log
-TRACE = False
+#debugrob: 
+#TRACE = False
+TRACE = True
```
Spaghet: Turn this off pls

PR #292, opened 2017-06-27
Rob trace es muy bueno

rbaynes:
Fix a few more things in the arduino handler:
- high / low water sensor pin assignments
- string bool value bug, changed by using float and comparison.

### DIFF
--- a/nodes/arduino_handler.py
+++ b/nodes/arduino_handler.py
@@ -29,8 +29,8 @@ sensor_csv_headers = (
     ("air_temperature", float),
     ("air_carbon_dioxide", float),
     ("water_temperature", float),
-    ("water_level_low", bool),
-    ("water_level_high", bool),
+    ("water_level_low", float),
+    ("water_level_high", float),
     ("water_potential_hydrogen", float),
     ("water_electrical_conductivity", float)
 )
@@ -203,15 +203,14 @@ def light_intensity_red_callback(msg): # float 0~1
     actuator_state["light_intensity_red"] = command
 
 
-# The water level sensor is HIGH when dry, which gets passed through the
-# linear_controller node, which takes the sensor value */measured
-# and passes it as */commanded. We should set the pump_5_water_1 to HIGH when
-# we receive a True here. I want to deprecate this with something more
-# feedback loop oriented: https://github.com/OpenAgInitiative/openag_brain/issues/270
-def water_level_high_callback(msg): # Bool
-    command = msg.data
-    if command:
+def water_level_high_callback(msg): # float 1 / 0
+    command = float(msg.data)
+    trace('arduino_handler debugrob water_level_high_callback=>%s< ', command)
+    # if the high water level is 1, turn the pump on (until it reaches 0)
+    if command >= 1.0:
         actuator_state["pump_5_water_1"] = True
+        trace('arduino_handler debugrob X SETTING pump_5_water_1=%d', actuator_state["pump_5_water_1"])
+
 
 CALLBACKS = {
     "air_temperature":          air_temperature_callback,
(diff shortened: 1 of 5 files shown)


nodes/arduino_handler.py
```diff
@@ -202,15 +203,15 @@ def light_intensity_red_callback(msg): # float 0~1
...
 
-# The water level sensor is HIGH when dry, which gets passed through the
-# linear_controller node, which takes the sensor value */measured
-# and passes it as */commanded. We should set the pump_5_water_1 to HIGH when
-# we receive a True here. I want to deprecate this with something more
-# feedback loop oriented: https://github.com/OpenAgInitiative/openag_brain/issues/270
-def water_level_high_callback(msg): # Bool
-    command = msg.data
-    if command:
+#PID controller sends us float values ranging from very small to almost 1.0.
```
Spaghet: For future reference the water level is not connected to a PID controller but it's connected to a direct_controller in the current launchfile configuration.

nodes/arduino_handler.py
```diff
@@ -306,15 +307,10 @@ def process_message(line):
...
-    rospy.init_node('arduino_handler')
-
-    publisher_rate_hz = rospy.get_param("~publisher_rate_hz", 1)
-    serial_rate_hz = rospy.get_param("~serial_rate_hz", 1)
-    baud_rate = rospy.get_param("~baud_rate", 115200)
-
+def connect_serial(scon=None):
     serial_rate = rospy.Rate(serial_rate_hz)
-    timeout_s = 1 / serial_rate_hz
+    to = 1 / serial_rate_hz
```
Spaghet: I prefer being verbose as possible when possible. "to" could mean anything, so we should use timeout or keep it at timeout_s to indicate what unit it is in. to -> timeout -> timeout in seconds is 2 inferences that you need to make, making your code that much harder to read later

nodes/arduino_handler.py
```diff
@@ -362,9 +371,12 @@ def process_message(line):
...
             actuator_state["chiller_compressor_1"]
         ).encode('utf-8')
-        serial_connection.write(message)
-        serial_connection.flush()
-        #rospy.logwarn("arduino_handler serial write {} bytes: >{}<".format(len(message),message.replace('\n','')))
+        try:
+            serial_connection.write(message)
+            serial_connection.flush()
+            trace('arduino_handler serial write %d bytes: >%s<', len(message), message.replace('\n',''))
+        except Exception as e:
```
Spaghet: I think the exception you get here is `serial.serial_util.SerialException` we should be explicit as possible whenever possible.

nodes/arduino_handler.py
```diff
@@ -362,9 +371,12 @@ def process_message(line):
...
+            serial_connection = connect_serial()
 
         # Read from arduino
         buf = serial_connection.readline()
```
Spaghet: Also this is from a previous PR but we should check the buffer length and 1st letter inside of process_message since it is already responsible for reading a verifying the message it gets.
I think the second check ("bad bytes") is already done within process message.
The 0 length is also caught by process_message, but I agree we should give it a different error case than the short read.

src/openag_brain/recipe_interpreters.py
```diff
@@ -28,7 +28,7 @@ def interpret_simple_recipe(recipe, start_time, now_time):
     """
     _id = recipe["_id"]
     operations = recipe["operations"]
-    rospy.logdebug(operations)
+    #rospy.logdebug(operations)
```
Spaghet: 😬

nodes/recipe_handler.py
```diff
@@ -155,8 +155,8 @@ def start_recipe_service(self, data, start_time=None):
             return False, "\"{}\" does not reference a valid "\
             "recipe".format(recipe_id)
 
-        trace("recipe_handler: PUBLISHERS=%s", PUBLISHERS)
-        trace("recipe_handler: recipe=%s", recipe)
+        #trace("recipe_handler: PUBLISHERS=%s", PUBLISHERS)
```
Spaghet: 🤢

nodes/arduino_handler.py
Spaghet: *linear_controller

Issue #291, opened 2017-06-26
Arduino: "invalid command: 100"


Spaghet:
Sometimes, the command that was meant to go to the nutrient pumps will go into the pwm_actuator for the LEDs, causing it to error out.

This is caused by the split function in src.ino not properly checking the comma count and should be easy to fix.

Issue #290, opened 2017-06-26
Arduino initializes forever


Spaghet:
Sometimes, the Arduino will send an empty message, then a message that says "1, MHZ16, 1, Initializing" and sway between these 2 states for a while.

Spaghet:
I think this doesn't happen anymore but I'll leave it in since we didn't strictly look into what was causing this.

Issue #236, opened 2017-05-16
Arduinio: If I2C crashes, Wire library can not recover (does not try to reconnect)


rwdavis513:
If we remain with the arduino, we likely need to replace the wire library or update the code so it will automatically reconnect

wsnook:
Are you familiar with I2C, Wire, and writing Arduino firmware? From your description, I think you may have the wrong idea about what I2C and Wire do and how they work.

Wire is the standard I2C library for Arduino, Photon, and maybe other similar platforms. It provides low level primitives for reading and writing on the I2C bus. Wire function calls return error codes about things like whether there was a timeout (according to the I2C bus timing specs) or how many bytes were read from a slave device. Wire isn't supposed to reconnect to anything or to do any sort of error recovery--that's not part of its job.

I2C is a serial bus specification that uses open-drain signaling. The bus doesn't have state, so it can't crash. But, it can have electrical problems like too much capacitance in the wiring, inadequately sized pull-up resistors, a device connected to the bus that isn't using I2C signaling (e.g. an Atlas sensor in TTL serial mode), or a mismatch between the bus master's clock speed and the min/max clock speeds supported by the wiring and the slave devices.

For talking to I2C slave devices with an Arduino, protocol state and error recovery are the responsibility of device specific drivers that--if you're lucky--follow the manufacturer's advice in the sensor datasheet for things like timing and how to configure registers. Typically, the sensor manufacturers provide datasheets, and, in the context of Arduino, the drivers come from third parties who sell the sensors on a breakout board for hobbyists. The libraries written for hobbyists tend to have minimal error handling, and they don't always follow the datasheet.

gordonbrander:
@jakerye 👆

Spaghet:
The cause of the entire Arduino crashing when an I2C device is faulty seems to be that the library will hang when there is the wrong length reply given by a I2C device.
One approach we could take is to add in an error case for the Wire library to handle when the received message in `endTransmission()` is of an insufficient bit length and throw an error after a specific timeout. It does handle error cases like NACKs so it doesn't seem unreasonable to handle this case in Wire as well.

Reconnecting isn't particularly necessary but it would be good to handle this error case.

wsnook:
@Spaghet Interesting... perhaps I'm wrong about Wire. How are you diagnosing that the library hangs or that an I2C device gives the wrong length of reply? Do you have error messages?

Spaghet:
[...]

PR #293, opened 2017-06-28
Hotfix invalid command

Spaghet:
This branch is just a couple bug fixes for the arduino_handler/arduino sketch combination. 

I also did some refactoring but I rescind my removal of printf/print/logdebug/warn.

### DIFF
--- a/nodes/arduino_handler.py
+++ b/nodes/arduino_handler.py
@@ -300,17 +300,11 @@ def process_message(line):
     except IndexError:
         message = "arduino_handler: Partial message: >{}<".format(line)
         rospy.logwarn(message)
-        if serial_connection is not None:
-            serial_connection.close()
-            serial_connection.open()
         return message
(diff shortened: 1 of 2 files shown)


Spaghet:
This fixes #291. 

#290 doesn't happen anymore, but I didn't do anything in particular to remove it so I assume it still exists on some machines. It is probably related to #236

nodes/arduino_handler.py
```diff
@@ -293,39 +300,37 @@ def process_message(line):
...
+    timeout_s = 1 / serial_rate_hz
     baud_rate = rospy.get_param("~baud_rate", 115200)
 
-    # below: mutables (gasp!)
     # Initialize the serial connection
     path = "/dev/serial/by-id"
     port = None
-    if not os.path.exists(path):
-      raise IOError("No serial device found on system in {}".format(path))
+    while port is None:
```
rbaynes: nice retry loop

PR #294, opened 2017-06-29
Fix serial read buffer overflow and enable openag_brain to work with couchdb with authentication.

goruck:
I am integrating an Alexa voice interface with the PFC which requires code running in an AWS Lambda instance to access the PFC's apis over the Internet. Therefore for security purposes its desirable to end the couchdb admin party. This PR enables openag_brain to work with couchdb with basic authentication enables. See https://github.com/goruck/foodcomputer for context and additional details.

### DIFF
--- a/nodes/arduino_handler.py
+++ b/nodes/arduino_handler.py
@@ -331,7 +331,6 @@ def connect_serial(serial_connection=None):
             rospy.logwarn(e)
             rospy.sleep(0.2) #seconds
 
-
 if __name__ == '__main__':
     if TRACE:
         rospy.init_node('arduino_handler', log_level=rospy.DEBUG)
@@ -349,6 +348,9 @@ if __name__ == '__main__':
     actuator_state["water_circulation_pump_1"] = True
 
     publish_time = ros_next(publisher_rate_hz)
+
+    temp_buf = ""
+
     while not rospy.is_shutdown():
 
         # Generate the message for the current state (csv headers below):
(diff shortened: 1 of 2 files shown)


Spaghet:
Thanks for taking time to look into this! Also, your Alexa integration stuff looks amazing.

I think this looks ok, but I'm not familiar enough with this codebase right now to make a valid assessment.
It looks like you would need to modify the install script so that we set up cert keys in the install script if we were to push these changes, and they might also affect the opeang_ui.

I agree it's important to make these changes so I will look into this immediately after validating the arduino driver/firmware fix we've been working on.

goruck:
Hi,

Thanks for the feedback and review.

Note that the normal http url still works without authentication if couchdb
still is in admin party mode. So for local use and for those who don't care
about secured access over the Internet nothing needs to be changed. I
tested both http and https with and without authentication with these code
changes and it seems to work fine.

My thinking was for the people that wanted to secure couchdb and run it
over https they would follow the standard couchdb instructions that I
summarized on my Alexa GitHub repo. This includes generating and placing
the keys and certs in the appropriate place. So I didn't see a need to make
it part of the install. But I'm happy to take a look at that. Just let me
know. However, I'm guessing when all the cloud db stuff comes on line
secure access becomes a must?

Keep up the amazing work!


On Jun 29, 2017 9:53 AM, "Rikuo Hasegawa" <notifications@github.com> wrote:

Thanks for taking time to look into this! Also, your Alexa integration
stuff looks amazing.

I think this looks ok, but I'm not familiar enough with this codebase right
now to make a valid assessment.
It looks like you would need to modify the install script so that we set up
cert keys in the install script if we were to push these changes, and they
might also affect the opeang_ui.

I agree it's important to make these changes so I will look into this
immediately after validating the arduino driver/firmware fix we've been
working on.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/OpenAgInitiative/openag_brain/pull/294#issuecomment-312027100>,
or mute the thread

PR #297, opened 2017-07-05
DoserPump Tuning

Spaghet:
I tuned the DoserPump dosage rates based on actual hardware tests.
I tried both pump 1 and 2 5 times each and took an average to get to the parameters.
The units are in `ml/hr`.


Spaghet:
The timeout where it gets angry if there hasn't been a recent command will still happen in this PR.

If you're talking about the PID timeout thing I'll send you the issue where I described that on gitter

PR #298, opened 2017-07-05
Air flush by volume

Spaghet:
This change changes the way recipes talk about the `air_flush` environmental variable(previously `air_flush_on`). We embed the CFM information of the air_flush fan into the firmware module, and write in the recipe the volume of air flush over time in `ft^3/min`. This change also deprecates the `air_flush_on` variable name and replaces it with `air_flush`.

### DIFF
--- a/firmware/src/src.ino
+++ b/firmware/src/src.ino
@@ -168,7 +168,7 @@ void actuatorLoop(){
   chiller_fan_1.set_cmd(str2bool(splitMessages[6]));              // BinaryActuator bool
   chiller_pump_1.set_cmd(str2bool(splitMessages[7]));             // BinaryActuator bool
   heater_core_2_1.set_cmd(str2bool(splitMessages[8]));            // BinaryActuator bool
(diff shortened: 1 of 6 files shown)


firmware/src/src.ino
```diff
@@ -168,7 +168,7 @@ void actuatorLoop(){
   chiller_fan_1.set_cmd(str2bool(splitMessages[6]));              // BinaryActuator bool
   chiller_pump_1.set_cmd(str2bool(splitMessages[7]));             // BinaryActuator bool
   heater_core_2_1.set_cmd(str2bool(splitMessages[8]));            // BinaryActuator bool
-  air_flush_1.set_cmd(str2bool(splitMessages[9]));                // BinaryActuator bool
+  air_flush_1.set_cmd(splitMessages[9].toFloat());                // BinaryActuator bool
```
rbaynes: comment should be something like:
```
//BinaryActuator float cfm 0 - 12ish?
```

PR #299, opened 2017-07-05
Ph loop

### LINKED ISSUE #285: Desired persists across recipes
If a recipe sets a desired value for a environmental variable, and then you run another recipe that doesn't set that environmental variable, the most recently broadcast desired value for that environmental variable remains to be commanded.

Spaghet:
I tuned the pH loop and fixed #285

### DIFF
--- a/launch/personal_food_computer_v2.launch
+++ b/launch/personal_food_computer_v2.launch
@@ -14,7 +14,7 @@
       <param name="lower_limit" value="-1" type="double"/>
       <param name="variable" value="water_potential_hydrogen" type="str"/>
       <param name="Kp" value="1" type="double"/>
-      <param name="Ki" value="0" type="double"/>
+      <param name="Ki" value="1" type="double"/>
       <param name="Kd" value="0" type="double"/>
     </node>
     <node pkg="openag_brain" type="topic_filter.py" name="topic_filter_1"/>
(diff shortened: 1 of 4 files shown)


launch/personal_food_computer_v2.launch
```diff
@@ -8,25 +8,25 @@
...
     <node pkg="openag_brain" type="pid.py" name="water_potential_hydrogen_controller_1">
-      <param name="Kd" value="0" type="double"/>
-      <param name="Ki" value="0" type="double"/>
       <param name="deadband_width" value="0.5" type="double"/>
       <param name="windup_limit" value="1000" type="double"/>
       <param name="upper_limit" value="1" type="double"/>
       <param name="lower_limit" value="-1" type="double"/>
       <param name="variable" value="water_potential_hydrogen" type="str"/>
       <param name="Kp" value="1" type="double"/>
+      <param name="Ki" value="1" type="double"/>
```
rbaynes: I'm just checking that it was intentional to change the integral value from zero to one here.

nodes/pid.py
```diff
@@ -110,6 +106,12 @@ def state_callback(item):
     def set_point_callback(item):
         pid.set_point = item.data
 
+    def recipe_end_callback(item):
```
rbaynes: Perhaps a comment to help the noobs like me?
```
# When we receive the recipe end message, reset this PID controller to its default values.  This disables the set point so the controller will just idle until it is set by a new recipe.
```

PR #308, opened 2017-07-10
fix reading empty messages - it was the serial port timing out

rbaynes:


nodes/arduino_handler.py
```diff
@@ -383,6 +383,10 @@ def connect_serial(serial_connection=None):
             trace('arduino_handler serial write %d bytes: >%s<', len(message), message.replace('\n',''))
             # Read
             buf = serial_connection.readline()
+            # handle reading nothing, happens when serial port times out
```
Spaghet: We already do this in process_message.

nodes/arduino_handler.py
```diff
@@ -309,7 +309,7 @@ def process_message(line):
 
 
 def connect_serial(serial_connection=None):
-    timeout_s = 1 / serial_rate_hz
+    timeout_s = 5 # 5 second serial port timeout
```
Spaghet: I don't like it but I can live with it

nodes/arduino_handler.py
rbaynes: OK, I will remove the check from here.  The other thing I was doing is sleeping a bit, but I don't think that matters.

nodes/arduino_handler.py
rbaynes: Instead of a hacky hard coded '5', how about we make the timeout 2x the rate (actually .5 the rate, so 2 secs).

rbaynes:
May prevent https://github.com/OpenAgInitiative/openag_brain/issues/305

nodes/arduino_handler.py
Spaghet: If we let process_message run, it will do the sleep at the end of the while loop

nodes/arduino_handler.py
Spaghet: `2 / serial_rate_hz` works for me

Issue #303, opened 2017-07-07
Commanded variables persist in arduino_handler


Spaghet:
If a environment variable isn't being commanded to, the arduino_handler keeps sending the most recent commanded state to the Arduino.

We should either implement a timeout or simply reset the message to the default state every loop.

Issue #310, opened 2017-07-10
stop_recipe does not emit recipe_end topic


Spaghet:
When we quit a recipe midway by calling the `/environments/environment_1/stop_recipe` service, it doesn't emit anything to the `/environments/environment_1/recipe_end/desired` topic. 

On a separate note, the UI seems to fail always when we call the stop_recipe service with status code 400 (bad request).

PR #311, opened 2017-07-10
[WIP] Bulk bugfix

Spaghet:
Working on a few different bugs.

recipe variables persisting through multiple recipes:
https://github.com/OpenAgInitiative/openag_brain/issues/303

Some nodes were publishing directly to var/measured so I made sure they publish to /raw so topic_filter can properly deal with it.

I'm going to flatten a few different bugs as well so don't review yet

### DIFF
--- a/nodes/arduino_handler.py
+++ b/nodes/arduino_handler.py
@@ -15,6 +15,7 @@ Usage (in ROS launchfile):
 import serial
 import rospy
 import os
+from collections import OrderedDict
 from std_msgs.msg import String
 from roslib.message import get_message_class
 from openag_brain.load_env_var_types import VariableInfo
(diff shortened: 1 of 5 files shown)


Spaghet:
https://github.com/OpenAgInitiative/openag_brain/issues/310

nodes/arduino_handler.py
```diff
@@ -210,12 +214,9 @@ def light_intensity_red_callback(msg): # float 0~1
...
 def water_level_high_callback(msg): # float 1 / 0
-    command = float(msg.data)
-    # if the high water level is > .5, turn the pump on
-    if command > 0.5:
-        actuator_state["pump_5_water_1"] = True
-    else:
-        actuator_state["pump_5_water_1"] = False
+    command = msg.data
+    # if the high water level is >= .5, turn the pump on
+    actuator_state["pump_5_water_1"] = command >= 0.5
```
rbaynes: Less readable, but understandable python.

Spaghet: I have a very strong aversion towards the use of `if True then True else False` statements..

PR #336, opened 2017-08-24
a little cleanup and remove a line that would cause an exception

rbaynes:
New serial_close() function and some cleanup.
One exception causing line removed.
Extending @goruck 's awesome fixes.

### DIFF
--- a/nodes/arduino_handler.py
+++ b/nodes/arduino_handler.py
@@ -314,13 +314,21 @@ def process_message(line):
         rospy.logwarn(message)
         return message
 
-def connect_serial(serial_connection=None):
+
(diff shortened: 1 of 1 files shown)


nodes/arduino_handler.py
```diff
@@ -392,7 +400,6 @@ def connect_serial(serial_connection=None):
         # Fix issue #328, sometimes serial_connection is None because of a 
         # serial port path / or error opening issue.
         if serial_connection is None:
-            serial_connection.close() # Forces DTS on reconnect to reset Arduino
```
rbaynes: since serial_connection is None here, this line would have thrown an Exception.

Issue #250, opened 2017-05-19
Docker install script complains about unset locale


Spaghet:
`scripts/install_docker.sh` fails when the locale is unset in the raspi.
We should figure out a way to set a default locale if the locale is unset in the raspberry pi in the `install_docker.sh` script.

wsnook:
I'm pretty sure there's an issue where gordon and I were discussing how to invoke raspi-config in non-interactive mode. I'll try to find a link to that stuff.

wsnook:
Found it. This was talking about enabling i2c, but I think the syntax could probably be adapted for the raspi-config's locale screen.

https://github.com/OpenAgInitiative/openag_brain/pull/159#issuecomment-290231712

Spaghet:
I'll give it a try

PR #338, opened 2017-08-27
Fix docker build

davoclavo:
## Fix Dockerfile

Currently the Dockerfile is broken due to changes to the install scripts. The Dockerfile now mostly relies on `scripts/install_dev`, and due to its dependency on resin's raspbian image, a couple of bits had to be modified.

- Create gpio group
- Give the user pi access to video group
- Changed virtualenv to be installed via pip

I think relying on `scripts/install_dev` is not the best way to solve this, but it works in the interim, and it keeps the docker and non-docker builds rely on the same installation script. The reason why I say is not the best solution is that then docker can't leverage caching builds between commands, so every time a change is made in `script/install_dev`, the whole step has to be re-built, which takes a long time. Also, changes to other files that `install_dev` relies on (like `init_dev`) probably won't be picked up in order to expire the cache, so could potentially cause troubles while debugging docker builds.

## Move couchdb related stuff out of `scripts/init_dev`

 - Compaction setting to `db_bootstrap/db_config.py`
 - Installation of couchdb to its own `scripts/install_db` script.

## Add docker-compose.yml

This file originally lived at the [openag_brain_docker_rpi repo](https://github.com/OpenAgInitiative/openag_brain_docker_rpi/blob/master/docker-compose.yml), but probably was lost in the transition to the monorepo. I think it is super useful to have it, and it is still mentioned a couple times in the docs.

**NOTE** - The [docker image in docker hub](https://hub.docker.com/r/openag/rpi_brain/) should be updated by one of the admins in order for [docker-compose to build from it](https://github.com/OpenAgInitiative/openag_brain/pull/338/commits/296d6d46e2c25adf44ef249229ee87c2963ed052#diff-4e5e90c6228fd48698d074241c2ba760R10).

----

## Testing

In order to test, make sure you have installed docker and docker-compose 

`./openag_brain/scripts/install_docker`

Then build the containers and start the application - This step will take a while, and you should be able to see progress in the logs.

`docker-compose up -d`

[Docker cheatsheet](https://blog.jez.io/2015/07/12/docker-tips-and-cheatsheet/)


scripts/install_pio
```diff
@@ -2,7 +2,7 @@
 #type platformio > /dev/null 2>&1 && echo "pio is already installed" && exit
 # We install platformio in a virtualenv, because platformio and rosserial
 # depend on different versions of pyserial.
-sudo apt-get install -y python-virtualenv
+sudo pip install virtualenv
```
Spaghet: Any particular reason for this change? Curious if it no longer works

scripts/init_dev
```diff
@@ -37,13 +37,8 @@ echo "Update apt"
 sudo apt-get update -q
 # sudo apt-get -y upgrade
 
-# Install couchdb (couchdb is not available via rosdep -- see below)
-echo "Install pip, couchdb and ROS bootstrapping tools"
-sudo apt-get install -y -q couchdb python-setuptools
-
-# Configure couchdb to do auto DB and view compaction
-# (or we will run out of disk space on a 32G SD card in a few weeks).
-curl -X PUT http://localhost:5984/_config/compactions/_default -d '"[{db_fragmentation, \"40%\"}, {view_fragmentation, \"30%\"}, {from, \"23:00\"}, {to, \"06:00\"}]"'
```
Spaghet: :)

Dockerfile
```diff
@@ -21,43 +21,28 @@ RUN apt-get update && \
     libboost-filesystem-dev \
     libxml2-dev \
     libgtest-dev
-# Tell apt to read from the ROS package repository
-RUN sudo sh -c 'echo "deb http://packages.ros.org/ros/ubuntu jessie main" > /etc/apt/sources.list.d/ros-latest.list'
-RUN wget https://raw.githubusercontent.com/ros/rosdistro/master/ros.key -O - | sudo apt-key add -
+
 RUN sudo locale-gen en_US.UTF-8
+ENV LANG=en_US.UTF-8
```
Spaghet: Fix for #250, much simpler than I expected :)

scripts/install_pio
davoclavo: When building using docker, I was getting the following errors:

```
...
/home/pi/catkin_ws/src/openag_brain/scripts/install_pio: line 6: virtualenv: command not found
/home/pi/catkin_ws/src/openag_brain/scripts/install_pio: line 7: /home/pi/pio_env/bin/activate: No such file or directory
```

And there wasn't any straightforward error regarding the `sudo apt-get install virtualenv`, so perhaps it silently fails (given that the execution gets to line 6), or perhaps the binary is just not available in the PATH? Not sure to be honest, but installing it via pip worked consistently 💯

Perhaps we can [add `set -e`to `install_pio`](https://github.com/OpenAgInitiative/openag_brain/blob/develop/scripts/install_dev#L6-L8) as well, so if the command fails the whole execution halts.

Spaghet: > Perhaps we can add set -eto install_pio as well, so if the command fails the whole execution halts.

Great idea.

> installing it via pip worked consistently 💯
I see. If @rbaynes can reproduce this it should be good to go.

rbaynes:
I'm setting up a clean raspbian SD card to test that this:
- doesn't mess up the developer build
- creates a docker instance that runa

rbaynes:
@Spaghet please re-review.  I made a few minor changes to make this work for the developer install.  (I'm still testing docker).  https://github.com/OpenAgInitiative/openag_brain/pull/338/commits/700efeed70bc13219bfbc9b5623efde1172bed72

scripts/install_dev
```diff
@@ -12,9 +12,9 @@ if [ ! -d ~/catkin_ws/src/openag_brain/ ] ; then
   echo "Location \"~/catkin_ws/src/openag_brain\" does not exist. openag_brain must be placed in \"~/catkin_ws/src\" for this script to work." && exit 1
 fi
 ~/catkin_ws/src/openag_brain/scripts/init_dev || exit 1
+~/catkin_ws/src/openag_brain/scripts/install_db || exit 1
```
davoclavo: My opinion is that couchdb related things shouldn't be part of `install_dev`, because the docker build is using docker compose to [provision the db in a separate container](https://github.com/OpenAgInitiative/openag_brain/pull/338/files#diff-4e5e90c6228fd48698d074241c2ba760R30), decoupling its management, and making it available in the brain container as `http://db:5984` (and to the outside world as well as `http://your-ip:5984`) 

If a single script installation is desired for non-docker builds, perhaps a new script that contained just the brain installation steps (something like `scripts/install_brain`) could be executed by both docker and the `install_dev` script (which would also run `install_db` to install couchdb)

rbaynes: Ah, good point, thanks David.  I don't ever use docker (which is probably obvious).    I will make a new "developer_setup" script and take that line out of install_dev.

